PR #4, opened 2020-11-02
 use case for taxonomic changes and collections data

sharifX:
A use case to help data curators track name changes and link rot.


hardistyar (approved):
Looks ok to me.

PR #23, opened 2021-07-20
add bird nest use case

sharifX:
A use case describing bird nest specimens, related eggs and anthropogenic materials found in the nest such as plastic water bottles.


hardistyar (approved):
I don't know that particleIdentification and particleDescription are the correct terms and note also that @smrgeoinfo in #24 has made some suggestions. Nevertheless, the use case is real and so we must confront it, especially the general issues of; i) representing relations between different DS; and ii) dealing with things that have multiple parts/features and those relations.

PR #44, opened 2023-07-17
Jupyter notebook for MAS tutorial

samleeflang:
A Jupyter notebook tutorial to explain what is required in a Machine Annotation Service.
This is an example, implementation can be found in: https://github.com/DiSSCo/demo-enrichment-service-image

The formatting in the PR looks strange as it just shows the text representation.
In the repo it should be able to show the cells and the code

### DIFF
--- a/machine-annotation-services/mas-tutorial.ipynb
+++ b/machine-annotation-services/mas-tutorial.ipynb
@@ -1,0 +1,273 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "source": [
+    "# Machine Annotation Service\n",
+    "\n",
+    "In this document we will show how you can build a machine annotation service (MAS) which fully integrates with the DiSSCo infrastructure.\n",
+    "When these steps are followed, DiSSCo will be able to deploy, scale and run your MAS and make it available for all relevant digital objects.\n",
+    "\n",
+    "In general, we expect a MAS to consists of three steps:\n",
+    "1. Receive Digital Object\n",
+    "2. Work with the object, this could be anything from calling external APIs to data cleaning\n",
+    "3. Publish Annotations\n",
+    "4. Publish code as container image\n",
+    "\n",
+    "This tutorial with help with steps 1 and 3.\n",
+    "Step 2 is up to the provider to implement.\n",
+    "\n",
+    "For receiving and publishing data DiSSCo uses Kafka.\n",
+    "Kafka helps with the decoupling as well as the automatically scaling the applications.\n",
+    "The MAS will receive a message on its own queue, will be triggered to consume the message and publishes its results also on a queue.\n",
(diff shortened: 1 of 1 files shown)


sharifX (changes requested):
shouldn't this be saved as .ipnyb? instead of .py? I tried to run it in locally but i needed python kafka module. also tried in google colab. it probably needs some kafka setup. it is not runnable in .py format. 

 1448         :return: tuple (host, port, afi), afi will be socket.AF_INET or socket.AF_INET6 or socket.AF_UNSPEC
   1449     """
-> 1450     host_and_port_str = host_and_port_str.strip()
   1451     if host_and_port_str.startswith('['):
   1452         af = socket.AF_INET6

AttributeError: 'NoneType' object has no attribute 'strip'

samleeflang:
> shouldn't this be saved as .ipnyb? instead of .py? I tried to run it in locally but i needed python kafka module. also tried in google colab. it probably needs some kafka setup. it is not runnable in .py format.
> 
> 1448 :return: tuple (host, port, afi), afi will be socket.AF_INET or socket.AF_INET6 or socket.AF_UNSPEC 1449 """ -> 1450 host_and_port_str = host_and_port_str.strip() 1451 if host_and_port_str.startswith('['): 1452 af = socket.AF_INET6
> 
> AttributeError: 'NoneType' object has no attribute 'strip'

Hmm it is saved as `.ipnyb` but is not runable as a notebook as it needs a Kafka connection.
It is more pieces of code commented with the markup cells.
Might be better to create a readme out of it, instead of a notebook, then it is clear that the code is not runnable.

wouteraddink (changes requested):
Good start, however I think the text need a little more explanation for new users about the general idea behind annotation services in DiSSCo and how the queuing works for people that do not know Kafka. Currently it seems to assume Kafka and Docker containers experience. I think Kafka knowledge should not be needed, Docker experience may be unavoidable. Also requirements and limitations are not yet well described. I do not understand the container image creation step, why should that be published in a public image registry, what do you mean with that? I guess you are talking here about docker container images not photo images which is a bit confusing.

samleeflang:
Would it makes sense to put that in here or write additional documentation for non-technical people. I wrote this with developers in mind. Shall I create a separate document with a more generic overview in Google docs which explains the ideas, requirements and limitations?

wouteraddink:
I think it can be done within this document, it only needs a few more sentences not pages of documentation. It should also be understandable for a researcher who occasionally writes some R or Python scripts to aid their work. Niels may be a good test person.

sharifX:
https://iiif.io/api/cookbook/ and https://iiif.io/api/cookbook/recipe/code/ have some nice examples/format for describing.

TomDijkema (changes requested):
Overall this sketches a good representation of the process needed to setup a MAS.
The reader will have to know about MAS, since there is no direct context given about the functionality of it.
Maybe add a link to other documentation that explains this further?

PR #52, opened 2023-10-10
Create annotation-schema-doc.md

sharifX:
created a schema doc using https://github.com/coveooss/json-schema-for-humans


sharifX:
we won't use this for now. we will create a short readme within each schema folder.

PR #53, opened 2023-10-12
Fix namespaces

southeo:
Few general fixes:
- Fix formatting according to DiSSCo practice
- fix missing namespace for "type"
- Add oa namespace to selector (should this be ods?)
- Capitalized constants "FieldValueSelector" and "ClassValueSelector"

### DIFF
--- a/data-model/annotations/schema/annotation.json
+++ b/data-model/annotations/schema/annotation.json
@@ -6,9 +6,7 @@
       "type": "string"
     },
     "ods:type": {
-      "enum": [
-        "Annotation"
-      ]
+      "const": "Annotation"
     },
     "version": {
       "type": "integer"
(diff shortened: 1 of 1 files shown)


data-model/annotations/schema/annotation.json
```diff
@@ -2,99 +2,207 @@
...
+        "ods:type": {
+          "type": "string"
+        },
+        "oa:selector": {
           "type": "object",
           "oneOf": [
             {
               "type": "object",
               "properties": {
-                "type": { "const": "fieldValueSelector" },
```
sharifX, line 24: this should be FieldValueSelector

data-model/annotations/schema/annotation.json
```diff
@@ -2,99 +2,207 @@
...
-              "required": ["ods:type", "ods:field"]
+              "required": [
+                "ods:type",
+                "ods:field"
+              ]
             },
             {
               "type": "object",
               "properties": {
-                "type": { "const": "classValueSelector" },
```
sharifX, line 32: This should be ClassValueSelector

data-model/annotations/schema/annotation.json
```diff
@@ -2,99 +2,207 @@
...
-        "ods:id": { "type": "string" },
-        "ods:type": { "type": "string" },
-        "selector": {
+        "ods:id": {
+          "type": "string"
+        },
+        "ods:type": {
+          "type": "string"
+        },
+        "oa:selector": {
```
sharifX, line 34: this is correct. selector is from oa namespace https://www.w3.org/ns/anno.jsonld

data-model/annotations/schema/annotation.json
```diff
@@ -2,99 +2,207 @@
   "$schema": "http://json-schema.org/draft-2020-12/schema#",
   "type": "object",
   "properties": {
-    "ods:id": { "type": "string" },
-    "ods:type": { "const": "Annotation" },
```
sharifX, line 6: this "type" is rdf:type based on w3c https://www.w3.org/ns/anno.jsonld. The other "types" are FDO Types so should ods.

sharifX (approved):

data-model/annotations/schema/annotation.json
```diff
@@ -2,99 +2,207 @@
...
-    "version": { "type": "integer" },
+    "ods:id": {
+      "type": "string"
+    },
+    "ods:type": {
+      "const": "Annotation"
+    },
+    "version": {
+      "type": "integer"
+    },
```
TomDijkema, line 13: I think @samleeflang mentioned during the planning that version can have 'ods' prefix

data-model/annotations/schema/annotation.json
southeo: so oa:type?

sharifX: No. rdf:type. anno.json refers to the rdf name space in the json ld context. https://www.w3.org/1999/02/22-rdf-syntax-ns#

TomDijkema (approved):

PR #54, opened 2023-10-16
add generated timestamp

southeo:
Add `generated` field. Fix "type" namespace.


TomDijkema (approved):

samleeflang (approved):
Generated timestamp was missing from the schema

PR #60, opened 2023-10-25
Update virtualcollections.json

sharifX:
First draft, json schema for virtualcollections.

This JSON schema defines a data structure for describing virtual collections
- **collection_type**: A string that specifies the type of collection, which can be either "Dynamic Virtual Collection" or "Static Virtual Collection."

- **id**: A unique identifier for the collection. Its format depends on the type of collection, and it's either a Handle for Dynamic Virtual Collections or a DOI for Static Virtual Collections.

- **filters**: An array of objects describing the filters selected for the collection. This is applicable to Dynamic Virtual Collections (DVC) and includes filter names and their corresponding values.

- **digital_specimens**: An array of objects representing the digital specimens included in the collection. This is applicable to Static Virtual Collections (SVC) and contains DOIs and the versions. 

- **is_private**: A boolean value indicating whether the collection is private (true) or public (false).

The schema enforces specific requirements based on the type of collection:

- For Dynamic Virtual Collections (DVC), the "filters" property is required.

- For Static Virtual Collections (SVC), the "digital_specimens" property is required.

### DIFF
--- a/data-model/virtualcollections/virtualcollections.json
+++ b/data-model/virtualcollections/virtualcollections.json
@@ -1,1 +1,88 @@
-
+{
+  "type": "object",
+  "properties": {
+    "collection_type": {
+      "type": "string",
+      "enum": ["Dynamic Virtual Collection", "Static Virtual Collection"]
+    },
+    "id": {
+      "description": "A unique identifier for the collection. It can be a Handle for DVC or a DOI for SVC.",
+      "anyOf": [
+        {
+          "if": {
+            "properties": { "collection_type": { "const": "Dynamic Virtual Collection" } }
+          },
+          "then": {
+            "type": "string"
+          }
+        },
+        {
+          "if": {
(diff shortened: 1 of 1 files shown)


samleeflang:
Couple of things:
- Would be good to add us as `Reviewers` so it gets on our radar. (I watch this repo so that how I noticed there was a new PR).
- Have you made this in line with the work we have already done last week for this (see user stories)? We already did most of this work last week and discussed how this would work.
- I don't think we should have one schema for both and Wouter indicated that static collections can be put on hold for now and  might not come. As per Slack conversation we would just call the Dynamic Collection, Virtual Collections and not have a distinction anymore between Dynamic and Static. Which we see as confusing for the user (if for us it is hard to keep them straight)
- It needs some additional metadata (created, creator, titile or name, etc...) see also the US for that.

sharifX:
i will update the schema and submit another PR.

PR #67, opened 2023-11-29
Added documentation to schema and small changes

samleeflang:
Added additional documentation:
- Descriptions with if possible linkages to the ontology
- Put things in constant when they appear to be constant
- Added examples so that it is clear what is expected
- Added enums when possible to limit the amount of possible values
- Added a comment on generated as this is not a valid oa term (not sure what to do yet)

Small changes
- `dcterms:reference` has become `dcterms:references` as this wasn't a valid dcterms term

### DIFF
--- a/data-model/annotations/schema/annotation.json
+++ b/data-model/annotations/schema/annotation.json
@@ -46,7 +46,7 @@
       "properties": {
         "ods:id": {
           "type": "string",
-          "description": "This is the handle of the target object. Valid targets are the Digital Specimen, Digital Media Object or another annotation.",
+          "description": "This is the PID of the target object. Valid targets are the Digital Specimen, Digital Media Object or another annotation.",
           "examples": [
             "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX",
             "https://doi.org/10.22/XXX-XXX-XXX"
@@ -74,7 +74,7 @@
                 },
                 "ods:field": {
                   "type": "string",
-                  "description": "The full jsonPath of the field being annotated.",
+                  "description": "The full jsonPath of the field being annotated. Following: https://goessner.net/articles/JsonPath/index.html#e2",
                   "examples": [
                     "$.occurrences[0].location.dwc:country"
                   ]
@@ -95,7 +95,7 @@
                 },
                 "oa:class": {
                   "type": "string",
-                  "description": "The full jsonPath of the class being annotated.",
+                  "description": "The full jsonPath of the class being annotated. Following: https://goessner.net/articles/JsonPath/index.html#e2",
                   "examples": [
                     "$.occurrences[0].location.georeference"
                   ]
@@ -153,7 +153,7 @@
                 "dcterms:conformsTo": {
                   "type": "string",
                   "constant": "https://ac.tdwg.org/termlist/#711-region-of-interest-vocabulary",
-                  "description": "http://purl.org/dc/terms/conformsTo"
+                  "description": "https://purl.org/dc/terms/conformsTo"
                 }
               },
               "required": [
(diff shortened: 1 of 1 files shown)


data-model/annotations/schema/annotation.json
```diff
@@ -3,45 +3,77 @@
   "type": "object",
```
TomDijkema, line 3: Suggestion to add a title to every JSON schema on the top level.
Gave the other models in my local copy also titles so the FE can use that to automatically generate type files (for the naming).

TomDijkema (approved):
Important changes, we need to be keen on the target type since the value of an annotation is always a string, the target type (field or class) will indicate if it hosts a literal string or a stringified object.

sharifX (approved):

data-model/annotations/schema/annotation.json
```diff
@@ -1,47 +1,83 @@
...
+          "description": "This is the handle of the target object. Valid targets are the Digital Specimen, Digital Media Object or another annotation.",
+          "examples": [
+            "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX",
+            "https://doi.org/10.22/XXX-XXX-XXX"
+          ]
         },
         "ods:type": {
-          "type": "string"
+          "type": "string",
+          "description": "This is the handle to the type of the target object.",
```
southeo, line 57: Maybe change handle to PID?

data-model/annotations/schema/annotation.json
```diff
@@ -161,27 +246,48 @@
...
-      "format": "date-time"
+      "format": "date-time",
+      "description": "Filled if the annotation was archived. The date and time when the annotation was archived"
     },
     "oa:generated": {
       "type": "string",
-      "format": "date-time"
+      "format": "date-time",
+      "description": "The date and time when the annotation was generated. http://purl.org/dc/terms/issued",
+      "$comment": "Be aware this is not a valid oa field. This is called generated within W3C annotation model but actually maps to dcterms:issued. We kept it generated to keep in line with W3C annotation standard."
```
southeo, line 261: If it's not a valid oa:field, should it be in the oa namespace? should we move it to ods?

data-model/annotations/schema/annotation.json
```diff
@@ -161,27 +246,48 @@
     },
     "dcterms:created": {
       "type": "string",
-      "format": "date-time"
+      "format": "date-time",
+      "description": "The date and time when the annotation was created. http://purl.org/dc/terms/created"
```
southeo, line 250: https

southeo (commented):
few comments

data-model/annotations/schema/annotation.json
samleeflang: We discussed this, but we don't have a solution. We could use the dc:terms to which it refers, but then we might lose the W3C compliancy. I don't think we should create our own term. That is why I put in the comment here, it needs another looks to check how to do this right.

data-model/annotations/schema/annotation.json
```diff
@@ -161,27 +246,48 @@
...
     },
     "as:generator": {
       "type": "object",
+      "description": "Object containing information on who generated the object. Generated is here seen as who stored/indexed the object. In most case this will be the annotation-processing-service",
       "properties": {
-        "ods:id": {
-          "type": "string"
+        "ods:type": {
+          "type": "string",
+          "description": "The type of the creator.",
```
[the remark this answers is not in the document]
samleeflang: We have created a vocab for those yet. It will be the handle of the types of DS, DM and annotation but those types are not yet fixed. Left it String when it would probably change. We can fix those to enums later

data-model/annotations/schema/annotation.json
samleeflang: Done

data-model/annotations/schema/annotation.json
```diff
@@ -193,13 +299,24 @@
...
+            3
+          ]
         },
         "schema.org:ratingValue": {
-          "type": "number"
+          "type": "number",
+          "description": "The rating for the content. https://schema.org/ratingValue",
+          "examples": [
+            4.5
+          ]
```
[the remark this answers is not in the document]
samleeflang: Done

samleeflang:
@southeo I should have fixed all your comments. The once I couldn't / won't fix I tried to answer as best as possible.

southeo (approved):
:1st_place_medal:

southeo:
All convos should be resolved!

PR #69, opened 2023-12-12
Feature/schema library

southeo:
- Adds deployment for schemas to be posted at schema.dissco.tech
- Adds schemas for base fdo profiles (DOI and Handle Kernels)
- Include versioning in file system

https://github.com/DiSSCo/dissco-core-deployment/pull/31


data-model/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+COPY ./annotations/0.1.0/schema/annotation.json schema-root/schemas/annotations/0.1.0/annotation.json
+
+COPY ./fdo-profiles/0.1.0/digital-specimen/schema/ schema-root/schemas/fdo-profiles/0.1.0/
+COPY ./fdo-profiles/0.1.0/doi-kernel/schema/ schema-root/schemas/fdo-profiles/0.1.0/
+COPY ./fdo-profiles/0.1.0/handle-kernel/schema/ schema-root/schemas/fdo-profiles/0.1.0/
+
+COPY ./specimens-and-media/0.1.0/digital-specimens/schema schema-root/schemas/digital-specimens/0.1.0/
+COPY ./specimens-and-media/0.1.0/shared-models/schema schema-root/schemas/shared-models/0.1.0/
+COPY ./specimens-and-media/0.1.0/digital-media-object/schema schema-root/schemas/digital-media-object/0.1.0/
```
TomDijkema, line 13: I feel like this might be a lot of work to go over all the versions when the schemas get updated.
Is there a way you could automatically include these files by the latest version and dynamically make the routes?

data-model/fdo-profiles/0.1.0/digital-specimen/schema/digital-specimen.json
```diff
@@ -0,0 +1,101 @@
...
+    "description": "Highest-level terms identifying the fundamentals of the activities, in which context the objects in the collection were collected.",
+    "$comment": "idx = 208"
+  },
+  "topicDomain": {
+    "enum": [],
+    "description": "",
+    "$comment": "idx = 209"
+  },
+
+  "": {
```
samleeflang, line 89: Don't think this is correct?

data-model/fdo-profiles/0.1.0/digital-specimen/schema/digital-specimen.json
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+  },
+  "topicOrigin": {
+    "enum": ["Natural", "Human-made", "Mixed origin"],
+    "description": "Highest-level terms identifying the fundamentals of the activities, in which context the objects in the collection were collected.",
+    "$comment": "idx = 208"
+  },
+  "topicDomain": {
+    "enum": [],
```
samleeflang, line 84: I think we got the enum for it in the digital-specimen opends schema

data-model/nginx/nginx.conf
```diff
@@ -0,0 +1,17 @@
+worker_processes 4;
+
+events {
```
samleeflang, line 3: Do we need this or can it be removed?

data-model/nginx/nginx.conf
```diff
@@ -0,0 +1,17 @@
+worker_processes 4;
```
samleeflang, line 1: What is the reason to put it on 4?

samleeflang (commented):
Very nice! Great that the FDO records are added as well :tada: 
Have a couple questions and comments

data-model/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM nginx:stable-alpine3.17-slim
+
+COPY ./nginx/nginx.conf /etc/nginx/nginx.conf
+
+COPY ./annotations/0.1.0/schema/annotation.json schema-root/schemas/annotations/0.1.0/annotation.json
```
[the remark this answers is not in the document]
southeo: My thoughts on versioning: these objects are not going to be versioned at the same rate, i.e. a release for a new FDO profile will be independent of annotations/specimens. 

If there's an update in the annotation model but not in the specimen model, the file structure might look like:

```
| --- 0.1.0
         | --- annotations/annotations.json
         | --- specimens/identifications.json
         | --- specimens/occurrences.json
         | --- media/digital-entity.json 
| --0.2.0 
         | --- annotations/annotations.json
```
Would make it harder to find the latest version for a given schema. If annotation model is up to 3.2.1 and specimens are still at 1.3.2...how would you know where to look? We could move all the schemas together, even if there's no changes, but that defeats the purpose of versioning imo. 

Long-winded way of saying I think it makes sense for versioning to be where it is.

data-model/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+COPY ./fdo-profiles/0.1.0/digital-specimen/schema/ schema-root/schemas/fdo-profiles/0.1.0/
+COPY ./fdo-profiles/0.1.0/doi-kernel/schema/ schema-root/schemas/fdo-profiles/0.1.0/
+COPY ./fdo-profiles/0.1.0/handle-kernel/schema/ schema-root/schemas/fdo-profiles/0.1.0/
+
+COPY ./specimens-and-media/0.1.0/digital-specimens/schema schema-root/schemas/digital-specimens/0.1.0/
```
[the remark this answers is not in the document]
southeo: (See comment on versioning for the beginning of my slow descent into madness) 

Specimens, media, and shared models are intrinsically linked, and I think it makes sense for their versioning to be under one folder. If something in shared-models changes, the references in digital-entity and the specimen schemas would need to be updated. 

Separate directories:

```
| --- specimens 
    | --- 0.1.0
         | --- occurrences.json
         | --- identifications.json
    | --- 0.2.0 
         | --- occurrences.json
         | --- identifications.json
| --- media
     | --- 0.1.0
         | --- digital-entity.json
     | --- 0.2.0 
         | --- digital-entity.json                            
| --- shared-models
     | --- 0.1.0
         | --- agent.json
     | --- 0.2.0 
         | --- agent.json
```

If we group these three directories under one release, it looks cleaner (or as clean as it can): 

```
| --- specimens-and-media
   | --- 0.1.0
      | --- specimens
         | --- occurrences.json
         | --- identifications.json
      | --- media
         | --- digital-entity.json
      | --- shared-models
          | --- agent.jsons
   | --- 0.2.0
      | --- specimens
         | --- occurrences.json
         | --- identifications.json
      | --- media
         | --- digital-entity.json
      | --- shared-models
          | --- agent.json
```
It's also easier in making sure nothing is left behind in a version release.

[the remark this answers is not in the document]
southeo: The trouble is, if we update agent.json, we're inadvertently updating both media and specimen models. I think the file system should reflect this.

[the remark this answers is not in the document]
southeo: I have been thinking about this A LOT over the last few days.

data-model/Dockerfile
southeo: Yeah exactly. The trouble with serving static content :upside_down_face:

data-model/fdo-profiles/0.1.0/digital-specimen/schema/digital-specimen.json
southeo: Meant to put this in an upcoming PR, will remove for now

data-model/nginx/nginx.conf
southeo: All the other nginx files we have have 4 so I just went with it. 

[From this nginx blog post](https://www.nginx.com/blog/tuning-nginx/)
- 1 process can handle 512 connections by default
- recommended to have 1 process per core
- Setting this this to `auto` will achieve that (better for auto-scaling clusters)

data-model/nginx/nginx.conf
southeo: I think you need it with an http block. Without it, the container crashes: 

```
[emerg] 1#1: no "events" section in configuration
nginx: [emerg] no "events" section in configuration
```

data-model/Dockerfile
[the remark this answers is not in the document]
samleeflang: ok sounds good

data-model/Dockerfile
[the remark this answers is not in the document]
samleeflang: Ok ok, yes and a better name than specimen-and-media? It doesn't look that great in the url. Can we call it `digitalobject` or something, not perfect (as everything is a digitalobject) but shorter and not with the `-and-` in between there?

data-model/nginx/nginx.conf
samleeflang: Ok wasn't sure, thanks for checking

data-model/nginx/nginx.conf
samleeflang: :+1: thanks for looking into this. Do you think we should change it in other place as well? Feel free to make a Jira issue for it

southeo: https://naturalis.atlassian.net/browse/DD-865

data-model/Dockerfile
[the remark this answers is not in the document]
southeo: done, also updated the references in the jsonschemas

samleeflang (approved):

PR #73, opened 2023-12-15
Run as nonroot

southeo:
Run nginx as nonroot, using [nginx/unprivledged](https://github.com/nginxinc/docker-nginx-unprivileged)

From documentation: 

> This repo contains a series of Dockerfiles to create an NGINX Docker image that runs NGINX as a non root, unprivileged user

I'm not sure how this will interact with the deployment's "run as nonroot" directive. We may need to remove that line in the deployment if we get issues. 

Change in nginx.conf is a fix to allow the unprivledged container to run ([docker hub](https://hub.docker.com/r/nginxinc/nginx-unprivileged) describes the fix)

### DIFF
--- a/data-model/Dockerfile
+++ b/data-model/Dockerfile
@@ -1,7 +1,5 @@
 FROM nginxinc/nginx-unprivileged:alpine3.18-slim
 
-RUN
-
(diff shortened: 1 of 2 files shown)


samleeflang (approved):
:clinking_glasses:

PR #79, opened 2024-01-16
Add job id to annotation model

southeo:
Job ID will facilitate tracking batch annotations


sharifX (approved):
jobid is needed to track batch annotation. so this is good.

TomDijkema (approved):

PR #61, opened 2023-11-03
Second updated version of virtualcollections.json

sharifX:
- This JSON schema consolidates the structure of Virtual Collections, eliminating the previous distinction between Dynamic and Static collections.
- It introduces collection properties and capabilities based on the Research Data Alliance (RDA) recommendations on Collection which is also adopted by the FDO Forum. More info here: http://rdacollectionswg.github.io/apidocs/#/ By simplifying the collection model and removing the Dynamic vs. Static differentiation, it aims to reduce potential user confusion.

PR #81, opened 2024-02-05
updated virtualcollections schema, align with LatimerCore

sharifX:
This updated virtual collection JSON schema has specific additions aligned with the LatimerCore (ltc) schema. The schema now incorporates the ltc:ObjectGroup class, encompassing properties: ltc:conditionsOfAccess, ltc:collectionName, ltc:description, and ltc:isKnownToContainTypes.  The descriptions included in the schema are taken from the LatimerCore documentation. The comments from https://github.com/DiSSCo/openDS/pull/61 has been addressed. The "capabilities" attributes suggest by RDA has been removed. Ltc also has ltc:RecordLevel based on RDA recommendation "The machine-actionable information profile for the collection description digital object." ods:type should be able map to that.


data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+          "description": "A short title that summarizes the collection objects contained within the ObjectGroup."
+        },
+        "ltc:description": {
+          "type": "string",
+          "examples": ["A virtual collection representing all butterfly specimens in South Holland"],
+          "description": "A free text description or narrative about the collection."
+        },
+        "ltc:isKnownToContainTypes": {
+          "type": "array",
+          "description": "Flag property to indicate that the collection is known to include type specimens. `true` - this collection contains types; `false` - this collections does not contain types.",
```
southeo, line 62: Why is this an array of booleans? 
This might be a costly flag to check. Remember, virtual collections are like saved searches, so they're dynamic. If we have a VC with 10 million items, are we supposed to check regularly if the collection contains a type specimen? 
@samleeflang thoughts?

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+          "items": {
+            "type": "boolean",
+            "examples": ["true"],
+            "description": "Type known to be contained."
+          }
+        }
+      },
+      "description": "An intentionally grouped set of objects with one or more common characteristics."
+    },
+    "ltc:MeasurementOrFact": {
```
southeo, line 72: How do these terms relate back to the concept of virtual collections? What kind of measurement did you have in mind?

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+        },
+        "dwc:measurementValue": {
+          "type": "number",
+          "examples": [100],
+          "description": "The numerical value of the measurement."
+        }
+      },
+      "description": "A measurement of or fact about a class within the standard, or a relationship between the ObjectGroup and an associated class."
+    },
+    "ods:search_filters": {
```
southeo, line 88: change to `ods:searchFilters`.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+      "type": "array",
+      "description": "Search filters selected for the virtual collection.",
+      "items": {
+        "type": "object",
+        "properties": {
+          "filter_name": {
```
southeo, line 94: change to `ods:filterName` (or just filterName - it should be in camelCase, but I'm not sure if we can claim this into ods)

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+            "type": "string",
+            "examples": ["c4df6c19a91c2daaeccd9628e3d3f196"],
+            "description": "Name or identifier/UUID for the filter."
```
southeo, line 97: Are we giving identifiers to filters?

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+        }
+      }
+    },
+    "ltc:basisOfScheme": {
+      "type": "string",
+      "examples": ["dataset for workshop"],
+      "description": "A summary of the basis or purpose for the LatimerCoreScheme. This property is intended to summarize the reason for grouping a number of ObjectGroups within the LatimerCoreScheme, and the purpose for which the data is intended to be used."
+    }
+  },
+  "required": ["ods:id", "ods:version","ltc:ObjectGroup"],
```
southeo, line 113: I don't see any required fields for ltc:objectGroup, is that intentional? Seems strange to require ltc:objectGroup but not set any required fields within it.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+        },
+        "foaf:name": {
+          "type": "string",
+          "examples": ["My Name"],
+          "description": "Name of the creator."
+        }
+      },
+      "description": "The creator of the collection."
+    },
+    "ltc:ObjectGroup": {
```
southeo, line 42: Should this be `ltc:objectGroup`? With lower case "o"? 
If so, it should also be changed in the "required" fields at the end of the schema.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+          },
+          "filter_value": {
+            "type": "array",
+            "examples": ["topicDiscipline=Zoology&midsLevel=1&country=Indonesia&hasMedia=1"],
+            "description": "Value or configuration of the filter."
+          }
+        }
+      }
+    },
+    "ltc:basisOfScheme": {
```
southeo, line 107: do we need this in our data model? we already have `ltc:description`

southeo (commented):
Thanks for tackling this. It's good to have latimer core linked to this data model. 
In the future, it might be easier to work with a clone the repo on a branch rather than a fork on your own github, so it's easier to keep your version up to-date :)

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,115 @@
...
+        }
+      },
+      "description": "The creator of the collection."
+    },
+    "ltc:ObjectGroup": {
+      "type": "object",
+      "properties": {
+        "ltc:conditionsOfAccess": {
+          "type": "boolean",
+          "examples": ["Open to the public"],
```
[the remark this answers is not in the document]
sharifX: ah, yes. this should not be boolean. it is a list. https://tdwg.github.io/ltc/terms/index.html#ObjectGroup_conditionsOfAccess

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: yes, it could be costly to check. But from the use case side, I see value. As a user, I want to find collections that has type specimens.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: I was debating this. Yes, description is enough but this from the ltc doc (https://tdwg.github.io/ltc/terms/index.html#LatimerCoreScheme_basisOfScheme) made me think that we could use it in a more FAIR/FDO Profile manner: "This property is intended to summarise the reason for grouping a number of ObjectGroups within the LatimerCoreScheme, and the purpose for which the data is intended to be used. This may also be reflected in the terms and metrics defined using the SchemeTerm and SchemeMeasurementOrFact classes respectively. Using this approach, standard and reusable profiles within the wider Latimer Core standard may be constructed for common collection descriptions use cases." We could implement this with controlled vocab. Examples given are these 
Accession
Inventory
Expedition
Digitisation planning
Collections assessment
Institution
Collections registry

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: fixed.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: number of object was one of examples given by ltc.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: it is with uppercase in ltc: https://tdwg.github.io/ltc/terms/index.html#ObjectGroup_ObjectGroup

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: let's keep this for now.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: removed this. value is the filter.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: added ltc:collectionName

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+    "ods:id": {
+      "type": "string",
+      "pattern": "^https:\\/\\/hdl\\.handle\\.net\\/20\\.5000\\.1025\\/.{3}-.{3}-.{3}$",
+      "examples": ["https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"],
+      "description": "PID of the object."
+    },
+   "rdf:type": {
+      "type": "string",
+      "const": "VirtualCollections",
+      "description": "The type is always Annotation. https://www.w3.org/TR/rdf12-schema/#ch_type"
```
samleeflang, line 14: This description needs to be updated as it still points to Annotation.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+  "properties": {
+    "ods:id": {
+      "type": "string",
+      "pattern": "^https:\\/\\/hdl\\.handle\\.net\\/20\\.5000\\.1025\\/.{3}-.{3}-.{3}$",
+      "examples": ["https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"],
+      "description": "PID of the object."
+    },
+   "rdf:type": {
+      "type": "string",
+      "const": "VirtualCollections",
```
samleeflang, line 13: Should this be "VirtualCollections" or should it point to the FDO type record as we do with other types?

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+        }
+      },
+      "description": "The creator of the collection."
+    },
+    "ltc:ObjectGroup": {
+      "type": "object",
+      "properties": {
+        "ltc:conditionsOfAccess": {
+          "type": "string",
+          "examples": ["Open to the public"],
```
samleeflang, line 47: Feels like this should be an enumeration. What choices do we have, ["Open", "Private"] maybe one or two others.

samleeflang: hmmm ok LC says you should point to a policy page when possible. Let's leave this a free text for now

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+          "examples": ["Arctiacaja1991Netherlands"],
+          "description": "https://tdwg.github.io/ltc/terms/index.html#ObjectGroup_collectionName A short title that summarizes the collection objects contained within the ObjectGroup."
+        },
+        "ltc:description": {
+          "type": "string",
+          "examples": ["A virtual collection representing all butterfly specimens in South Holland"],
+          "description": "https://tdwg.github.io/ltc/terms/index.html#ObjectGroup_description A free text description or narrative about the collection."
+        },
+        "ltc:isKnownToContainTypes": {
+          "type": "array",
```
samleeflang, line 61: This looks like a boolean field instead of an array field

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+          "description": "https://tdwg.github.io/ltc/terms/index.html#ObjectGroup_isKnownToContainTypes Flag property to indicate that the collection is known to include type specimens. `true` - this collection contains types; `false` - this collections does not contain types.",
+          "items": {
+            "type": "boolean",
+            "examples": ["true"],
+            "description": "Type known to be contained."
+          }
+        }
+      },
+      "description": "An intentionally grouped set of objects with one or more common characteristics."
```
samleeflang, line 70: This description doesn't seem to match the term

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+          "examples": ["Object Count"],
+          "description": "https://tdwg.github.io/ltc/terms/index.html#MeasurementOrFact_measurementType The nature of the measurement, fact, characteristic, or assertion."
+        },
+        "dwc:measurementValue": {
+          "type": "number",
+          "examples": [100],
+          "description": "https://tdwg.github.io/ltc/terms/index.html#MeasurementOrFact_measurementValue The numerical value of the measurement."
+        }
+      },
+      "description": "https://tdwg.github.io/ltc/terms/index.html#MeasurementOrFact_MeasurementOrFact A measurement of or fact about a class within the standard, or a relationship between the ObjectGroup and an associated class."
```
samleeflang, line 86: Maybe put this description directly after the term, it feels a bit detached at the moment.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+    },
+    "ltc:MeasurementOrFact": {
+      "type": "object",
+      "properties": {
+        "dwc:measurementType": {
+          "type": "string",
+          "examples": ["Object Count"],
+          "description": "https://tdwg.github.io/ltc/terms/index.html#MeasurementOrFact_measurementType The nature of the measurement, fact, characteristic, or assertion."
+        },
+        "dwc:measurementValue": {
```
samleeflang, line 80: If we do dwc:measurementValue we should definitely have dwc:measurementUnit in is well. The Unit indicates what the value represents, so whether it is a percentage or a count or in meters, etc... http://rs.tdwg.org/dwc/terms/measurementUnit

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
...
+        }
+      }
+    },
+    "ltc:basisOfScheme": {
+      "type": "string",
+      "examples": ["dataset for workshop"],
+      "description": "A summary of the basis or purpose for the LatimerCoreScheme. This property is intended to summarize the reason for grouping a number of ObjectGroups within the LatimerCoreScheme, and the purpose for which the data is intended to be used."
+    }
+  },
+  "required": ["ods:id", "ods:version","ltc:ObjectGroup, ltc:collectionName"],
```
samleeflang, line 108: I would make `rdf:type` also required

samleeflang, line 108: A quatation is missing after `"ltc:ObjectGroup,` should be `"ltc:ObjectGroup",`

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,110 @@
-
+{
+  "$schema": "http://json-schema.org/draft-2020-12/schema#",
```
samleeflang, line 2: Needs additional metadata id and a comment, see for example the digital specimen:
 "$id": "https://schemas.dissco.tech/schemas/digitalobjects/0.1.0/digital-specimens/digital-specimen.json",
 "$schema": "https://json-schema.org/draft/2020-12/schema",
 "$comment":"DigitalObject Version 0.1.0",

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
samleeflang, line 1: Could you add an example as well and make sure those validate against the schema with for example https://www.jsonschemavalidator.net/
Then we would be sure that we understand the schema correctly and that it is valid JSON.

samleeflang (commented):
Almost there, got a couple of additional things too look at.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: i think for our data model schema, we have been using string so let's keep this string now. but we need to align it for all the schema if we point to the FDO type. I think we do it in the FDO Profile schema.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: ok

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: yes, fixing this.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: it is the desc of the Objectgroup class taken from https://tdwg.github.io/ltc/terms/#ObjectGroup_ObjectGroup An intentionally grouped set of objects with one or more common characteristics.

samleeflang: Could we include the tdwg description as we did with the others? And put this next to the type

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
samleeflang: I don't think it is possible to set this and keep it in sync. What if new specimen come in or specimen are updated, we would have to recheck tens, hundreds, thousands? of virtual collections to see if this need an update as well... Might be best not to include this property.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,178 @@
...
+    },
+    "required": [
+        "rdf:type",
+        "ods:id",
+        "ods:version",
+        "ltc:ObjectGroup",
+        "ltc:collectionName"
+    ],
+    "additionalProperties": false,
+    "examples": [
```
samleeflang, line 145: Oeh I wouldn't put this in the file. It does bloat it a bit and isn't very finable. I also noticed that it doesn't validate. 
Two issues (says https://www.jsonschemavalidator.net/)
 Found 2 error(s)
Message:
Invalid type. Expected Boolean but got Array.
Schema path:
https://schemas.dissco.tech/schemas/digitalobjects/0.1.0/virtualcollections/virtualcollections.json#/properties/ltc:ObjectGroup/properties/ltc:isKnownToContainTypes/type
Message:
Required properties are missing from object: ltc:collectionName.
Schema path:
https://schemas.dissco.tech/schemas/digitalobjects/0.1.0/virtualcollections/virtualcollections.json#/required

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
samleeflang: Same for the count as well. We shouldn't put it in the object, but it should be calculated when it is displayed (so the filters are run). We can keep the fields the measurementOrFact fields as they are generic enough.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,178 @@
...
+  "ods:version": 1,
+  "dcterms:creator": {
+    "ods:type": "foaf:Person",
+    "foaf:name": "Mary Shelly"
+  },
+  "ltc:ObjectGroup": {
+    "ltc:conditionsOfAccess": "Open",
+    "collectionName": "ZoologyIndonesia",
+    "ltc:description": "A virtual collection representing Zoology specimens from Indonesia.",
+    "ltc:isKnownToContainTypes": [true]
```
samleeflang, line 158: This won't be in the actual object

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: removed the example. will add to the examaple folder.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,138 @@
...
+        },
+        "ltc:basisOfScheme": {
+            "type": "string",
+            "description": "A summary of the basis or purpose for the LatimerCoreScheme. This property is intended to summarize the reason for grouping a number of ObjectGroups within the LatimerCoreScheme, and the purpose for which the data is intended to be used."
+        },
+        "examples": [
+            "dataset for workshop"
+        ]
+    },
+    "required": [
```
samleeflang, line 130: I think searchFilters needs to be required.

samleeflang, line 130: Would probably make creator required as well

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,138 @@
...
+        "ods:searchFilters": {
+            "type": "array",
+            "description": "Search filters selected for the virtual collection.",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "filterValue": {
+                        "type": "array",
+                        "examples": [
+                            "topicDiscipline=Zoology&midsLevel=1&country=Indonesia&hasMedia=1"
```
samleeflang, line 115: This looks strange. I think we will probably change it to make sure what we need.
Not sure what the array is doing at the moment as everything is in 1 value.
I had more something like this in mind:
```
[ 
  { 
    "filterKey": "country",
    "filterValue": ["Netherlands", "Belgium"]
  },
  {
     "filterKey": "midsLevel",
     "filterValue": ["1"] // Keep filterValue an array of Strings
  }
]
```
So we would have an array of filterKey which is a string and filterValue which is an array of strings.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,138 @@
...
+            "const": "VirtualCollections"
+        },
+        "ods:version": {
+            "type": "integer",
+            "description": "The version of the object.",
+            "examples": [
+                1
+            ]
+        },
+        "dcterms:creator": {
```
samleeflang, line 27: In the annotation model we have two fields as required:
```
      "required": [
        "ods:id",
        "ods:type"
      ]
```

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -1 +1,138 @@
...
+                "ltc:description": {
+                    "type": "string",
+                    "description": "https://tdwg.github.io/ltc/terms/index.html#ObjectGroup_description A free text description or narrative about the collection.",
+                    "examples": [
+                        "A virtual collection representing all butterfly specimens in South Holland"
+                    ]
+                }
+            }
+        },
+        "ltc:MeasurementOrFact": {
```
samleeflang, line 79: I would say if you add a MeasurementOrFact all three need to be required.

samleeflang (commented):
Could you take a look at the required fields?
Also the example for the filter doesn't look great.
I think we make it a bit more reusable and less specific by storing it as key-value pairs.

samleeflang (commented):
I think my last comments are still relevant. They were mainly about missing required fields and about the structure of the filter field. Could you check if you pushed all changes?

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: added

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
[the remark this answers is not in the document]
sharifX: added

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: added

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -101,36 +105,65 @@
...
+        }
     },
     "required": [
         "rdf:type",
         "ods:id",
         "ods:version",
+        "ods:searchFilters",
+        "dcterms:creator",
         "ltc:ObjectGroup",
         "ltc:collectionName"
```
samleeflang, line 168: lts:collectionName is not part of the top level object but of the ObjectGroup, if it needs to be required it should be required for that object.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
```diff
@@ -101,36 +105,65 @@
...
-        ]
+        }
     },
     "required": [
         "rdf:type",
         "ods:id",
         "ods:version",
+        "ods:searchFilters",
+        "dcterms:creator",
         "ltc:ObjectGroup",
```
samleeflang, line 167: Noticed there are inconsistencies in the use of the capital. Some objects has an uppercase others not. I think so far we did not make use of uppercase for objects

samleeflang (approved):
Had some minor comments but will approve as well so it can go through

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: Yes, for our terms we are consistent. we are using lowercase then camelcase (like ods:physicalSpecimenIdType) but the rest are borrowed verbatim so I don't want to change them.

data-model/virtualcollections/v.0.1.0/schemas/virtualcollections.json
sharifX: ok. good point. ltc:ObjectGroup does not require ltc:collectionName. So I will remove it from here. ltc:ObjectGroup only has one required element ltc:baseTypeOfObjectGroup ("High-level terms describing the fundamental nature of objects in the ObjectGroup.") but this is not relevant for us.

PR #89, opened 2024-04-29
source system profile

southeo:
Adds request and profile for source system pid records

note: If we use “allof”, we can’t generate code using our code gen library. Made a jira ticket to address other schemas to align them with this one

### DIFF
--- a/data-model/fdo-profiles/0.1.0/source-system/schema/source-system-request.json
+++ b/data-model/fdo-profiles/0.1.0/source-system/schema/source-system-request.json
@@ -1,0 +1,56 @@
+{
+  "$id": "https://schemas.dissco.tech/schemas/fdo-profiles/0.1.0/source-system-request.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "description": "Kernel from which all DiSSCo FAIR Digital Object profiles are derived",
+  "$comment":"FDO Profile Version 0.1.0",
+  "properties": {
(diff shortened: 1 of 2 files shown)


data-model/fdo-profiles/0.1.0/source-system/schema/source-system-request.json
```diff
@@ -0,0 +1,56 @@
...
+  "description": "Kernel from which all DiSSCo FAIR Digital Object profiles are derived",
+  "$comment":"FDO Profile Version 0.1.0",
+  "properties": {
+    "fdoProfile" : {
+      "type": "string",
+      "description": "PID to a machine readable description of the attributes in the FDO record",
+      "example": "https://hdl.handle.net/21.T11148/d8de0819e144e4096645",
+      "$comment": "idx = 1"
+    },
+    "digitalObjectType": {
```
TomDijkema, line 14: Shouldn't this be fdoType?

TomDijkema (commented):
I am missing a bit of context, are you updating the source system profile or this a completely new schema for the fdo of a source system?

data-model/fdo-profiles/0.1.0/source-system/schema/source-system-request.json
southeo: No, this is correct. It's based on the handle profile, which calls this field "digitalObjectType"

TomDijkema (approved):
Not really into these fdo profiles, but this looks ok!

PR #91, opened 2024-05-14
Feature/add fdo types

southeo:
Adds Schemas for
- Mapping
- Source System
- MAS

Reorganize our schemas a bit. Now we have 3 top level directories:
- fdo-profiles
- fdo-types
- annotations

The `annotations` directory contains non-fdo schemas, e.g. schemas for MAS providers to adhere to. For consistency, I've copied the annotation schema into the fdo-type directory as well. The id of the annotation Type schema points to the fdo directory now.


data-model/fdo-types/0.1.0/machine-annotation-services/schema/mas.json
```diff
@@ -0,0 +1,105 @@
...
+        "containerTag": {
+          "type": "string",
+          "description": "Tag of the image"
+        },
+        "targetDigitalObjectFilters": {
+          "type": "object",
+          "description": "Filters describing the criteria that must be met in order to apply the MAS. No filters implies the MAS may run on *any* digital object",
+          "$comment": "Needs to be defined"
+        },
+        "serviceDescription": {
```
samleeflang, line 38: Maybe a couple of descriptions for these, I could help as well if unclear

data-model/fdo-types/0.1.0/machine-annotation-services/schema/mas.json
```diff
@@ -0,0 +1,105 @@
...
+        },
+        "containerImage": {
+          "type": "string",
+          "description": "Image of the containerized application"
+        },
+        "containerTag": {
+          "type": "string",
+          "description": "Tag of the image"
+        },
+        "targetDigitalObjectFilters": {
```
samleeflang, line 33: An example would be nice as it is a bit of a complex object

samleeflang (commented):
Sorry a bit extra work, but some more descriptions and examples would be nice I think

sharifX (approved):
the new schemas and re-org looks good.

PR #93, opened 2024-05-28
Web batch annotations

southeo:
Schemas for batching web annotations

To batch request batch annotations through DiSSCover, front end sends an AnnotationEvent, which includes the annotation and batch metadata, adhering to the `annotation-event-for-web` schema. The contents of the annotation are not different from the non-batched annotations, but they are in an array. What's different is that we now require BatchMetadata, which contains search parameters for the batching functionality downstream. 

`Annotations` and `BatchMetadata` need to be a list containing exactly one item per field.

"place in Batch" is a constant set in the backend. tom, you only need to worry about the search parameters

### DIFF
--- a/data-model/annotations/0.1.0/examples/example-annotation-event-for-web.json
+++ b/data-model/annotations/0.1.0/examples/example-annotation-event-for-web.json
@@ -1,0 +1,36 @@
+{
+  "data": {
+    "type": "annotation",
+    "attributes": {
+      "annotations": [
+        {
+          "oa:motivation": "oa:commenting",
+          "oa:target": {
+            "ods:id": "https://hdl.handle.net/20.500.125/1111111",
+            "ods:type": "https://doi.org/21.T11148/894b1e6cad57e921764e",
+            "oa:selector": {
+              "ods:type": "ClassSelector",
+              "oa:class": "specimenName"
+            }
+          },
+          "oa:body": {
+            "ods:type": "body",
+            "oa:value": [],
+            "dcterms:reference": "reference",
+            "ods:score": 0.9
+          }
+        }
+      ],
+      "batchMetadata": [
+        {
+          "searchParams": [
+            {
+              "inputField": "field",
+              "inputValue": "value"
+            }
+          ]
+        }
+      ]
+    }
+  }
+}
--- a/data-model/annotations/0.1.0/schema/annotation-event-for-web.json
+++ b/data-model/annotations/0.1.0/schema/annotation-event-for-web.json
@@ -1,0 +1,166 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://schemas.dissco.tech/schemas/annotations/0.1.0/annotation-event-for-web.json",
+  "description": "Web batch annotation. For DiSSCo internal use only.",
+  "$comment": "Fewer fields are required in this schema because these requests are sent to the backend for processing.",
+  "properties": {
+    "annotation": {
+      "type": "object",
+      "properties": {
+        "ods:id": {
+          "type": "string"
+        },
+        "rdf:type": {
+          "const": "Annotation"
+        },
+        "oa:motivation": {
+          "enum": [
+            "ods:adding",
+            "oa:assessing",
+            "oa:editing",
+            "oa:commenting"
+          ]
+        },
+        "oa:motivatedBy": {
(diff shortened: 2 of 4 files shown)


data-model/annotations/0.1.0/schema/annotation-event-for-web.json
```diff
@@ -0,0 +1,189 @@
...
+                  }
+                },
+                "required": [
+                  "oa:motivation",
+                  "oa:target",
+                  "oa:body"
+                ]
+              },
+              "minItems": 1,
+              "maxItems": 1
```
TomDijkema, line 165: So there is only one annotation allowed per request?
What if there is MAS with annotations on multiple fields?

data-model/annotations/0.1.0/examples/example-annotation-event-for-web.json
```diff
@@ -0,0 +1,36 @@
...
+            "ods:score": 0.9
+          }
+        }
+      ],
+      "batchMetadata": [
+        {
+          "searchParams": [
+            {
+              "inputField": "field",
+              "inputValue": "value"
```
TomDijkema, line 29: What is input field and what is input value?

data-model/annotations/0.1.0/schema/annotation-event-for-web.json
southeo: This schema is specifically for requesting batch annotations from the Front End. you only send one annotation to the backend from DiSSCover.

[Event for MAS](https://github.com/DiSSCo/openDS/blob/master/data-model/annotations/0.1.0/schema/annotation-event.json). 

MASs are allowed to send multiple annotations, but the front end is not. (That could change, but we don't currently support it). 

batching works like:
1. **front end sends single annotation** and batch metadata, which defines the parameters for what other annotations to batch
1a. backend searches the parameters, returns total number of objects that would be annotated according to the batch metadata
1b. user confirms that number of objects should be annotated
2. processing service finds the annotated objects that meet that criteria, build annotations, then sends it back to itself to process the new batch result annotations

so the FE will only send one annotation, even in a batch request

data-model/annotations/0.1.0/examples/example-annotation-event-for-web.json
southeo: added examples, see the description of these two fields above:
input field: "JsonPath (mixed notation) of field used to base annotation reasoning on. Indexes must be wildcards"
input value: "Value stored at the field indicated in inputField"

data-model/annotations/0.1.0/schema/annotation-event-for-web.json
TomDijkema: Sounds like a good approach!

data-model/annotations/0.1.0/examples/example-annotation-event-for-web.json
southeo: value of the object. 

say you have:
```
inputField = occurrences[*].weather.temperature
inputValue = 20 C

```
Maybe you want to annotate the field `occurrences[*].weather.condition` with the value "that sounds like a nice day!"

So basically, input field/value is saying, "because this field is equal to this value. I am making my annotations thusly." and then we'd look for all specimen with `occurrences[*].weather.temperature = 20 C`, and make that annotation

southeo: Not quite. input field/value is not the same as the annotation target. **It defines the search criteria used to find the objects we want to make the batch annotations on.**

Having the search param:

```
searchParams: 
[{
  inputField: occurrences[*].weather.temperature
  inputValue: ""
}]

```

Would mean that we find all specimens where temperature is empty. Maybe then you make the following annotation on ALL specimens with an empty temperature field: 

```
annotation: {
   target: occurrences[*].weather.condition
   body: I wonder what the weather was like!
}
```

It may be possible that the annotation target field and the input field for the search params are the same. Like if you want to annotate a missing value or something. but it's helpful to think of them as completely separate concepts.

TomDijkema: Ah ok, so the search params really exist to find other specimens which should also be annotated within the batch.
That makes sense. The name convention of the properties is very plain so that's why I was wondering.

TomDijkema (approved):
Let's see this in action soon!

PR #102, opened 2024-06-27
Feature/event changes

samleeflang:
- Added fields for annotation status, these needs to capture who approved/rejected the annotation
- Fixes to the annotation object 
- Added examples for annotation object
- Changes to the createUpdateTombstoneEvent object so we can capture multiple agents


wouteraddink (commented):
waiting for the latest changes in annotationStatus

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
```diff
@@ -390,6 +389,56 @@
     "ods:batchID": {
       "type": "string",
       "description": "Internally generated PID to identify the batch the annotation was generated by"
+    },
+    "ods:mergingStatus": {
```
southeo, line 393: Why not capture if a decision is pending? If a machine wants to find if a decision has been made, it would have to do a null check. 
something like `ods:DecisionPendingForMerging`
`

southeo (approved):
:green_salad: 

One comment but I assume it's something that was already discussed

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
samleeflang: hmm yes you might be right, so when we create an annotation with motivation adding/editing we will fill this status with DecisionPending. We can then easily filter all annotations which require a decision

data-model/fdo-type/annotation/0.3.0/examples/tombstoned-annotation.json
```diff
@@ -0,0 +1,49 @@
...
+    "@id": "https://hdl.handle.net/20.5000.1025/ABC-DEF-GHI",
+    "@type": "as:Application",
+    "schema:name": "Annotation Generator Service"
+  },
+  "dcterms:issued": "2024-06-27T12:00:00Z",
+  "ods:TombstoneMetadata": {
+    "@type": "ods:Tombstone",
+    "ods:tombstonedDate": "2023-06-15T14:30:00Z",
+    "ods:tombstonedReason": "This object was made by mistake",
+    "ods:TombstonedByAgent": {
```
sharifX, line 43: TombstonedByAgent should lower case? -> tombstonedByAgent?

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
```diff
@@ -264,54 +271,13 @@
...
-            "Gbif linker service"
-          ]
-        }
-      },
-      "required": [
-        "@id",
-        "@type"
-      ],
-      "additionalProperties": false
+      "$ref": "https://schemas.dissco.tech/schemas/fdo-type/shared-model/0.3.0/agent.json"
```
sharifX, line 274: yes. I think is better to have the agent as a separate object.

data-model/fdo-type/shared-model/0.3.0/schema/tombstone-metadata.json
```diff
@@ -0,0 +1,67 @@
+{
```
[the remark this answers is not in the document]
southeo: Really happy we're talking about tombstoning, but we should think about the fdo profiles alongside these kinds of changes to the data model, or else we'll in for future headaches. :)

data-model/fdo-type/source-system/0.3.0/schema/source-system.json
```diff
@@ -61,20 +69,10 @@
         "2021-06-01T12:00:000Z"
       ]
     },
-    "ods:dateTombstoned": {
+    "schema:creator": {
       "type": "string",
```
southeo, line 73: can we use the `schema` namespace and then point to an `ods` object (agent)? This doesn't seem to me like it's an instance of `schema:creator`

data-model/fdo-type/shared-model/0.3.0/schema/citation.json
```diff
@@ -33,7 +33,8 @@
     },
     "dcterms:creator": {
```
southeo, line 34: if we're not using the dcterms definition anymore, is it still in the dcterms namepsace?

data-model/fdo-type/shared-model/0.3.0/schema/agent.json
```diff
@@ -7,61 +7,55 @@
   "properties": {
```
southeo, line 7: Throughout the data model, we call instances of this schema: 

- dcterms:creator
- as:generator
- schema:maintainer
- ods:MergingStateChangedBy
- schema:creator
- ods:hasAgent (arrays of agent)

In my opinion, only the latter is an appropriate usage. I don't think we can use other namespaces to point to an ods object.

We can capture the specifics of an agent, such as `MergingStateChangedBy` or `creator`, using the `schema:roleName` field in this object.

southeo: i have a few comments with this same idea but I can't find them, sorry for repeating myself here

southeo (changes requested):
:corn: 
Few things I think need to be addressed, if not changed, before this can be merged. (sorry for the big red x, i want to undo my previous approval...)

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
```diff
@@ -264,54 +271,13 @@
     "dcterms:creator": {
```
TomDijkema, line 271: Shouldn't this be called 'hasAgent' then?

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
samleeflang: Hmm it should be there, just checked and I can see.
```
      "enum": [
        "ods:DecisionPendingForMerging",
        "ods:RejectedForMerging",
        "ods:ApprovedForMerging"
      ],
```

data-model/fdo-type/annotation/0.3.0/examples/active-annotation-example.json
```diff
@@ -0,0 +1,37 @@
...
+    "foaf:name": "John Doe"
+  },
+  "dcterms:created": "2024-06-27T12:00:00Z",
+  "as:generator": {
+    "@id": "https://hdl.handle.net/20.5000.1025/ABC-DEF-GHI",
+    "@type": "as:Application",
+    "foaf:name": "Annotation Generator Service"
+  },
+  "dcterms:issued": "2024-06-27T12:00:00Z",
+  "ods:annotationStatus": "ods:Active"
```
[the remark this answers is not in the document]
samleeflang: No Active is the status of the FDO Object, Apporved, which is now `ods:ApprovedForMerging` is a status that the annotation is approved for merging into the original source object.

data-model/fdo-type/create-update-tombstone-event/0.3.0/examples/update-event-example.json
```diff
@@ -45,12 +58,23 @@
...
-    "prov:wasAttributedTo": "https://hdl.handle.net/20.5000.1025/ASC-WQW-AAU/2"
+    "prov:wasRevisionOf": "https://hdl.handle.net/20.5000.1025/9FS-W6R-L1L/1"
   },
-  "prov:Agent": {
-    "@id": "https://hdl.handle.net/20.5000.1025/ASC-WQW-AAU/2",
-    "@type": "prov:SoftwareAgent",
-    "foaf:name": "orchestration-backend"
-  }
+  "ods:hasProvAgent": [
+    {
```
TomDijkema, line 64: Why does this not point towards the agent object?

data-model/fdo-type/annotation/0.3.0/examples/approved-annotation-example.json
```diff
@@ -0,0 +1,44 @@
...
+  "as:generator": {
+    "@id": "https://hdl.handle.net/20.5000.1025/ABC-DEF-GHI",
+    "@type": "as:Application",
+    "foaf:name": "Annotation Generator Service"
+  },
+  "dcterms:issued": "2024-06-27T12:00:00Z",
+  "ods:batchID": "batch-1234",
+  "ods:annotationStatus": "ods:Approved",
+  "ods:decisionDate": "2024-06-28T12:00:00Z",
+  "ods:Decider": {
```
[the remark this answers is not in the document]
samleeflang: We don't use prov in the annotation object. prov-o is not a open set of terms such as schema's, dwc, dcterms but has a fixed structure. We cannot use a term from the namespace on any object.

data-model/fdo-type/annotation/0.3.0/examples/tombstoned-annotation.json
samleeflang: Yes that is what we discussed in as namingconvention, if we create new terms and they contain an object we capitalize them to indicate that it contains an object, in this case an agent.

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
samleeflang: :+1:

TomDijkema (commented):
Nice work! Two questions

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
```diff
@@ -27,6 +27,14 @@
         "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"
       ]
     },
+    "ods:status": {
```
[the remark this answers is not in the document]
samleeflang: Good point, I would be in favour of reducing this to DRAFT, ACTIVE, TOMBSTONE, TEST and put the reason why it was tombstoned in a different place. Now this is a bit mixed up, we have 5 statuses that indicate it was tombstoned.

data-model/fdo-type/shared-model/0.3.0/schema/tombstone-metadata.json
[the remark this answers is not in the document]
samleeflang: Yes, the term tombstonePids I didn't get and putting just a list of string in there without any context seemed wrong. In the FDO profile the status might indicate a relationship there, but I would be in favour of change the fdo profile at this point. The `ods:tombstoneReason` I will change to `tombstoneText` to keep in line.

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
southeo: yep I see it, looks good to me!

data-model/fdo-type/source-system/0.3.0/schema/source-system.json
samleeflang: Yes in the instance of `schema` we can actually do it as they leave it open-ended. In addition, the type of the object is not from the `ods` vocab. I you look in the agent you will see that we have five types defined there (where the `prov` types are only used in the provenance object).

data-model/fdo-type/shared-model/0.3.0/schema/agent.json
samleeflang: Ok, first thing is that we don't actually create objects of type `ods:Agent` but we create object with a couple types one being `schema:Person`. Schema is pretty open so we can use this in `schema:maintainer` and `schema:creator`. `as:generator` doesn't enforce a specific value type for the object https://www.w3.org/TR/activitystreams-vocabulary/#dfn-generator and neither does dcterms:creator (they recommend an URI, but doesn't say the value type is a string). So I actually don't think we go against any 'rules' when we do it this way.
The second point to include it all in one array and then use `schema:roleName` is a bit tricky as some of the ontologies we use do describe how the object need to look like. So the w3c web annotation model does indicate which terms need to be at what level to improve interoperability. So we can't just remove things such as dcterms:creator and as:generator from these objects.
Anyway that is how I interpreted the termsand when/what to reuse. Might be good if we could find an expert on this to have them check the model and see if we made any mistakes.

data-model/fdo-type/create-update-tombstone-event/0.3.0/examples/update-event-example.json
samleeflang: It does, this is a valid object. The agent-object is just a fixed set of properties we use for all places where we refer to agents.

TomDijkema: My bad its an example :0

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
wouteraddink: That would work, It looks like as end status for the annotation we really are capturing the mergin decision status, We could then also rename ods:mergingStatus to ods:mergingDecisionStatus with as enum: ods:pending, ods:rejected, ods:approved.

I wonder if we need an extra state for the object ods:status:merged? That may make it also easier for the front-end to decide whether to show an annotation. By default we may not want to show merged annotations with the digital specimen anymore. Or should they be tombstoned after merging? I think they are not active anymore as they cannot be further decided on for merging and should not be commented on further.  I guess annotations that were rejected for merging,  would also be tombstoned.

samleeflang: I agree, let's simplify, and call it `ods:mergingDecisionStatus`. I will push the changes in a moment.

I would say they get tombstoned, the annotation is no longer necessary as it has now become part of the object. After merging we can tombstone the annotation based on the `ods:mergingDecisionStatus` somebody can check if it was approved or rejected.

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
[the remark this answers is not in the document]
wouteraddink: These are valid pid statuses but yes I we could reduce the number of pid statuses by adding a TOMBSTONED status and have the rest as tombstone reasons. RETIRED is a status where you do not want to show a tombstone at all, as if the pid was never there. Think of a lawsuit situation where someone requests a pid to be no longer accessible. But for our use cases I think it is a bit hypothetical and we could just show a tombstone in all cases. Failed is if the pid string was created, but something went wrong with creating the pid metadata. it should not be tombstoned but pid creation should be re-attempted by the pid infrastructure. Depending on how the pid infrastructure was created we may not actually need it, Sou would know better.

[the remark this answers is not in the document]
wouteraddink: I think we did not do it because it is not really the pid which gets tombstoned but the object that the pid refers to.

data-model/fdo-type/shared-model/0.3.0/schema/tombstone-metadata.json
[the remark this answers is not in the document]
wouteraddink: tombstonePids is an array of pids to show to the user where he can go when the object is tombstoned. the context can then put in the tombstoneText and follows partly from pidStatus but I think that could be improved.

data-model/fdo-type/shared-model/0.3.0/schema/agent.json
wouteraddink: it sounds ok to me. Otherwise we end up with having to use ods terms instead and decisions that do not improve interoperability.

wouteraddink (changes requested):
I would add an example or a rejected annotation, as it is now not clear if that would change the ods:status (see my earlier comments). I assume this will remain ods:Active. 

Added fields for annotation status look good and also the changes to the createUpdateTombstoneEvent

data-model/fdo-type/annotation/0.3.0/schema/annotation.json
wouteraddink: ok if they get tombstoned then the example needs to be changed as well.

samleeflang:
Added annotation rejected example and updated the approved annotation example to display status = Tombstone

wouteraddink (approved):

TomDijkema (approved):
Very nice!

data-model/fdo-type/shared-model/0.3.0/schema/agent.json
southeo: @samleeflang  can you document this discussion somewhere, maybe in release notes? Otherwise sounds great

data-model/fdo-type/data-mapping/0.3.0/examples/tombstoned-data-mapping-example.json
```diff
@@ -0,0 +1,45 @@
+{
+  "@id": "https://hdl.handle.net/20.5000.1025/ABC-DEF-GHI",
+  "@type": "ods:DataMapping",
+  "ods:ID": "https://hdl.handle.net/20.5000.1025/ABC-DEF-GHI",
+  "ods:type": "https://doi.org/10.15468/1a2b3c",
+  "ods:status": "ods:Tombstone",
```
sharifX, line 6: should this be "Tombstoned"?

data-model/fdo-type/data-mapping/0.3.0/schema/data-mapping.json
```diff
@@ -34,6 +34,14 @@
         "https://doi.org/10.15468/1a2b3c"
       ]
     },
+    "ods:status": {
+      "enum": [
+        "ods:Draft",
+        "ods:Active",
+        "ods:Tombstone"
```
sharifX, line 41: Tombstoned?

data-model/fdo-type/digital-media/0.3.0/examples/tombstoned-digital-media-example.json
```diff
@@ -0,0 +1,34 @@
+{
+  "@id": "https://doi.org/10.22/ABC-DEF-GHI",
+  "@type": "ods:DigitalMedia",
+  "ods:ID": "https://doi.org/10.22/ABC-DEF-GHI",
+  "ods:version": 1,
+  "ods:status": "ods:Tombstoned",
```
sharifX, line 6: Here this is ok.

southeo (approved):
need to think about fdo profiles but that's beyond the scope.

sharifX:
ods:Tombstone is ok to be consistent with the rest of the ods status (Active, Draft)

data-model/fdo-type/digital-media/0.3.0/examples/tombstoned-digital-media-example.json
samleeflang: Fixed this example

PR #110, opened 2024-07-05
Add modified timestamp to objects

samleeflang:
As discussed during our last refinement.
Modified will contain the timestamp of the version

## Added modified timestamp to all (modifiable) Digital Objects
For the DigitalSpecimen and DigitalMedia objects we added a `dcterms:modified` property.
This will be taken over from the source system on first ingestion or, if not present, it will be equal to the `dcterms:created` (which has been renamed from `ods:created`) property.
Each time the object is updated, the `dcterms:modified` property should be updated to the current time (`dcterms:created` will remain the time of first creation).
Because we could take over the first value from the source system, `dcterms:modified` is a string, and we won't enforce a date-time format.
For the other Digital Objects, we will use `schema:dateModified`, which is more in line with the other terms.
It will be equal to `schema:dateCreated` when the object is first created.
On updates the `schema:dateModified` should be updated to the current time, the `schema:dateCreated` should remain the time of first creation.


TomDijkema (approved):
This is very good to have! :+1:

sharifX (approved):
Looks good. The justification to mix "dcterms:modified" and "schema:dateModified" gives us flexibility as not all the source data we are ingesting incorporates proper ISO date-time format.

wouteraddink (approved):
Looks good, no further comments.

PR #125, opened 2024-08-06
update regex

southeo:
update regex for fdo profile annotation: target pid
- this matches production dois and test/sandbox dois, which have an alphabetic prefix, not a numeric one


TomDijkema (approved):
:kangaroo:

PR #131, opened 2024-08-15
mas request wrapper

southeo:
MAS create/update requests now may contain environmental variables and secrets
[See PR on Orchestration Service](https://github.com/DiSSCo/dissco-core-orchestration-backend/pull/46)


samleeflang (approved):
Wonder if it isn't overkill but on the other hand this is what we need and it is optional

samleeflang:
Might have missed but maybe an example might help, not blocking though

PR #136, opened 2024-08-30
Add sourceSystemName and other minor changed

samleeflang:
- Adding SourceSystemName to openDS because we need it when we create annotations and CreateUpdateTombstoneEvents. The SourceSystem is often the agent of these actions (creating new specimen for example) we want to store the identifier (PID), a human readable name for the system and the type (as:Application) for the agent.
- This will also help aggregation, for which we also proposed this change. You would now be able to filter on a specific sourcesystem (Dataset in GBIF).
https://naturalis.atlassian.net/browse/DD-1327

Fixed required fields in dataMapping
Also some minor example updates to be more in line with reality.

### DIFF
--- a/data-model/fdo-type/source-system/0.3.0/schema/source-system.json
+++ b/data-model/fdo-type/source-system/0.3.0/schema/source-system.json
@@ -8,7 +8,7 @@
     "@id": {
       "type": "string",
       "description": "The unique identifier (handle) of the Source System object",
-      "pattern": "^https://hdl\\.handle\\.net/[\\w.]+/(.){3}-(.){3}-(.){3}",
+      "pattern": "^https:\/\/hdl\\.handle\\.net\/[\\w.]+\/(.){3}-(.){3}-(.){3}",
       "examples": [
         "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"
       ]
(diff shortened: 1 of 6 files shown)


data-model/fdo-type/source-system/0.3.0/schema/source-system.json
```diff
@@ -8,7 +8,7 @@
     "@id": {
       "type": "string",
       "description": "The unique identifier (handle) of the Source System object",
-      "pattern": "^https:\/\/hdl\\.handle\\.net\/[\\w.]+\/(.){3}-(.){3}-(.){3}",
+      "pattern": "^https://hdl\\.handle\\.net/[\\w.]+/(.){3}-(.){3}-(.){3}",
```
southeo, line 11: Was this a mistake? I think you need to escape these `/`

Also, if you're changing the pattern, i think it should be `\w+` not `[\\w.]+`

southeo (commented):
just a regex thing

data-model/fdo-type/source-system/0.3.0/schema/source-system.json
samleeflang: Reverted regex. I tested but you need the `.` in there to allow dots in the prefix part. Tested with https://regexr.com/ and this should be alright.

southeo: Right right, looks good

southeo (approved):
:game_die:

TomDijkema (approved):

PR #137, opened 2024-09-02
An agent name can not always be guaranteed

samleeflang:
Discussed with team, an agent name can not always be guaranteed.
This attribute can not be set to a required field.


TomDijkema (approved):
Better to have it, but I get that's not possible all the times (during processes)

southeo (approved):

PR #138, opened 2024-09-03
1.0 fdo profile

southeo:
Introduced 1.0.0 Version of FDO Profiles according to finalization discussion

**Handle Objects**
- Handle-kernel
- Annotation
- Data-mapping
- Machine-annotation-service
- Source-system

**DOI Objects**
- Doi-kernel
- Digital-specimen
- Digital-media

**Request vs normal schema?**
A "request schema" is used to validate requests to the handle API. Terms in these schemas are terms that may or must be determined by a client. terms not in the request schemas are generated internally by the handle api.


TomDijkema (approved):
:deer

data-model/fdo-profile/digital-media/1.0.0/schema/digital-media-request-attributes.json
```diff
@@ -0,0 +1,86 @@
...
+      "description": "Additional URLs to be appended to the 10320/loc field"
+    },
+    "referentName": {
+      "type": "string",
+      "description": "A generic name for the type of object that the DOI refers to. This is different from digitalObjectType that points to a specific type, e.g. there can be different types of digital specimens that each have a slightly different metadata schema because they describe a different kind of specimen, like a botanical versus a geological specimen.",
+      "example": "digital specimen"
+    },
+    "mediaHost": {
+      "type": "string",
+      "description": "ROR or, in absence of a ROR, Wikidata Qnumber for the administrative organisation responsible for hosting the digital media object. Note that this can be DiSSCo also (use DiSSCo QID in that case). Notehe organisation may use an external repository like morphobank to host the object.",
```
sharifX, line 21: typo fix. "note the.."

sharifX (changes requested):

data-model/fdo-profile/annotation/1.0.0/examples/example-annotation-request.json
wouteraddink, line 1: I would add a targetTypeName with the name of the targetType (in this case DigitalSpecimen) to follow the policy to use an identifier in combination with a name where possible.

wouteraddink: same for type, I would add typeName

wouteraddink: is a MD5 hash resistant enough to collisions?

data-model/fdo-profile/annotation/1.0.0/examples/example-annotation.json
wouteraddink, line 1: add VIEW=JSON to 10320/loc for consistency (although we do not currently have a HTML view for the annotation)

data-model/fdo-profile/annotation/1.0.0/schema/annotation.json
wouteraddink, line 1: update idx6 example with 10.3535 prefix, add VIEW=JSON to10320/loc

data-model/fdo-profile/data-mapping/1.0.0/schema/data-mapping.json
wouteraddink, line 1: I am missing issuedForAgent and issuedForAgentName

data-model/fdo-profile/digital-specimen/1.0.0/schema/digital-specimen-request-attributes.json
wouteraddink, line 1: typo in topicCategory, should be Microfungi not Mircofungi

wouteraddink: for topicDiscipline we need to add Archeology at some point (can leave it out now since not supported yet in DiSSCover)

wouteraddink (changes requested):

data-model/fdo-profile/annotation/1.0.0/examples/example-annotation-request.json
southeo: MD5 is definitely resistant enough to random collisions. 
- There's a 1/(2^128) chance of any 2 random messages colliding. 
- If we have 2^64 messages, there is a 50% chance of there being at least one collision (see [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem))
     - That's 1.8x10^19 annotations to get to that point. If we're at that many annotations, we have a few other technical issues to address first!

data-model/fdo-profile/data-mapping/1.0.0/schema/data-mapping.json
southeo: these terms were moved to only refer to the DOI profiles only

sharifX (approved):
looks good. a small comment.

data-model/fdo-profile/annotation/1.0.0/examples/example-annotation.json
wouteraddink: it is now implemented as ID=JSON but the ID should not change, a VIEW parameter should be added. ID attributes are given an integer in the Handle documentation, I am not sure that is actually required but people using Handles may expect an integer there so I would keep the ID integers. They may be handy if there are for example multiple URLS with VIEW=JSON, to display one of them with preference using a weight (e.g. one is served in Chine, the other in Europe and for traffic coming from Europe you may prefer the one served from a location in Europe).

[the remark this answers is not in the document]
wouteraddink: now pidIssuerName is changed to the ROR, but the ROR should be in pidIssuer and pidIssuerName should be DiSSCo

data-model/fdo-profile/digital-media/1.0.0/schema/digital-media-request-attributes.json
wouteraddink, line 1: dcterms:license was changed to licenseName, but filled in the example with a URL pointing to the license, for example https://spdx.org/licenses/CC0-1.0.json.  Following our policy to include both a name and an identifier, it may be best to have both a term 'licenseURL' for e.g. "https://spdx.org/licenses/CC0-1.0.json" and a term 'licenceName' for e.g. "Creative Commons Zero v1.0 Universal". In SPDX the licenseID is "CC0-1.0" but since we use the URL as the ID I think we do not need to include that.

data-model/fdo-profile/digital-media/1.0.0/schema/digital-media.json
wouteraddink, line 1: In the description you can remove "(use DiSSCo QID in that case)" since DiSSCo now has a ROR

wouteraddink (changes requested):
Almost there! a few requests for changes still, see comments

data-model/fdo-profile/annotation/1.0.0/examples/example-annotation.json
southeo: Hi Wouter,

Do you have additional documentation on how to implement this? based on the [handle.net documentation](https://www.handle.net/overviews/handle_type_10320_loc.html), there are the following selection methods defined for the 10320/loc field:
- locatt, which allows a client to select a location based on id (e.g. `20.5555/1234?locatt=id:0`)
- weight
- country

Nothing about selecting based on a views

Fortunately, there's nowhere in the documentation I could find that indicates the id must be an integer, so I implemented the proposed functionality using id instead. User-provided locations are given an integer identifier.

data-model/fdo-profile/digital-media/1.0.0/schema/digital-media-request-attributes.json
southeo: Exactly. The changes you describe were made in the digital media fdo profile (digital-media.json), but this is the schema for the fdo request made to the handle API. In the processing service, we're trying to map an open string field dcterms:license to these 2 fields. the handle API looks at the license coming in from the processing service, and determines the license it's resolvable or not.  

if it's resolvable, the handle api populates licenseUrl and licenseName with the resolved name. If it's not resolvable, the handle api populates just "licenseName". 

This structure exists based on the assumption that a client can't determine whether the license is resolvable or not. However, it's possible down the road some client has more structured data. for that reason, I'm adding 3 license terms: license (for when you don't know what license you have), and licenseUrl and licenseName for when you can distinguish between the two.

data-model/fdo-profile/annotation/1.0.0/examples/example-annotation.json
southeo: I'm not 100% sure if the handle system supports that. In [their documentation](https://www.handle.net/overviews/handle_type_10320_loc.html), they say: 

> The currently defined selection methods are:
> - Locatt (select by id)
> -  country
> - weight

However, they also say: 

> The attributes for the set of locations, as well as each location entry in the set, are open-ended to allow for future capabilities.

So it might support it, it might not. Will need to test. If it fails, using the id this way should work. There's nothing in the documentation that says that the id needs to be an integer.

wouteraddink: ok, how I read the documentation you should be able to add any attribute you like. If it does not work then please use the id instead but having both attributes would be more flexible.

wouteraddink (approved):

PR #142, opened 2024-09-09
mas env and secrets

southeo:
- Remove MAS request wrapper
- MAS environment and secrets are now part of the MAS itself


data-model/fdo-type/machine-annotation-service/0.3.0/schema/machine-annotation-service-secret.json
```diff
@@ -0,0 +1,49 @@
...
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "The name of an environmental variable stored remotely",
+      "examples": [
+        "database.password",
+        "keycloak.secret"
+      ]
+    },
+    "secretKeyRef": {
```
samleeflang, line 17: I wonder if this is necessary. I know we had it initially but then I was also wondering. Wouldn't it be possible to make this easier and have them just supply a list of key's (so the names) we then assume that the key is in the mas-secrets store and the key is something we generate for example?

samleeflang (commented):
Question about the secrets

data-model/fdo-type/machine-annotation-service/0.3.0/schema/machine-annotation-service-secret.json
southeo: was wondering the same. i don't think it's necessary right now, but maybe if we wind up having multiple secret stores down the road it could be useful to have it like this? can change it for now.

samleeflang (approved):
:lacrosse:

PR #143, opened 2024-09-09
align tombstone request with tombstone metadata

southeo:
updates tombstone request for handle api
We don't make pojos from this schema, it's just for completeness


data-model/fdo-profile/tombstone/1.0.0/tombstone-request.json
```diff
@@ -0,0 +1,73 @@
...
+                        "ods:isDuplicateOf"
+                      ]
+                    }
+                  },
+                  "additionalProperties": false
+                },
+                "additionalProperties": false,
+                "required": [
+                  "tombstoneText"
+                ]
```
samleeflang, line 52: There is no tombstoneText here? Maybe `pid` should be required at least, so if a relatedPid is there we at least need the pid

data-model/fdo-profile/tombstone/1.0.0/tombstone-request.json
```diff
@@ -0,0 +1,73 @@
...
+              }
+            }
+          },
+          "additionalProperties": false,
+          "required": [
+            "tombstoneText"
```
samleeflang, line 58: Not sure if we will always have a tombstoneText, although we could probably always put something in, even if it is not that informative

samleeflang (commented):
:open_file_folder:

data-model/fdo-profile/tombstone/1.0.0/tombstone-request.json
southeo: tombstoneText is the first field in `attributes`
also, the pid is provided in the `id` field

data-model/fdo-profile/tombstone/1.0.0/tombstone-request.json
southeo: We can discuss if it still needs to be required for the pid profile on our fdo meeting, but tombstone text is required for [ods tombstone attributes](https://schemas.dissco.tech/schemas/fdo-type/shared-model/0.3.0/tombstone-metadata.json), so we'll always have something

samleeflang: ah ok, good.

samleeflang (approved):
:womans_hat:

PR #145, opened 2024-09-09
add max records

southeo:
align with orchestration service


samleeflang (approved):
Oh yeah thanks forgot this one

PR #177, opened 2024-09-11
Fix utf 8 encoding and set default content-type to json

samleeflang:
- Content-type to application/json
- Force set encoding to UTF-8


southeo (approved):
# :palm_tree:

PR #196, opened 2024-09-18
openDS term data schema

sharifX:
A JSON schema to describe the openDS terms as FDO.
The first version of this schema is focused on creating the basic structure of the object. Feedback from [RFC](https://docs.google.com/document/d/1m6Dzz2fOy3ZCLL9De8gHKH0Pw3TWY3Az-9pSyDAxbyY/edit) will be incorporated in different stages. There are a few decisions regarding RDF, JSON-LD we have to make.

### DIFF
--- a/data-model/fdo-type/term/term.json
+++ b/data-model/fdo-type/term/term.json
@@ -1,0 +1,195 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://schemas.dissco.tech/schemas/fdo-type/term/latest/term.json",
+  "title": "DiSSCo openDS Terms Schema",
+  "description": "A schema for defining standard terms in openDS.",
+  "type": "object",
+  "properties": {
+    "@id": {
+      "type": "string",
+      "description": "The unique identifier (handle) of the term object",
+      "pattern": "^https://hdl\\.handle\\.net/[\\w.]+/(.){3}-(.){3}-(.){3}",
+      "examples": [
+        "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"
+      ]
+    },
+    "@type": {
+      "type": "string",
+      "description": "The type of the object, in this case ods:Term",
+      "const": "ods:Term",
+      "examples": [
+        "ods:Term"
+      ]
+    },
+    "rdf:type": {
+      "type": "object",
+      "properties": {
+        "@id": {
+          "type": "string",
+          "enum": [
+            "http://www.w3.org/1999/02/22-rdf-syntax-ns#Property",
+            "http://www.w3.org/2000/01/rdf-schema#Class",
+            "http://www.w3.org/2002/07/owl#Class",
+            "http://www.w3.org/2004/02/skos/core#Concept"
+          ]
+        }
+      }
+    },
+    "dcterms:identifier": {
+      "type": "string",
+      "description": "Handle of the Term",
+      "pattern": "^https://hdl\\.handle\\.net/[\\w.]+/(.){3}-(.){3}-(.){3}",
+      "examples": [
+        "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"
+      ]
+    },
+    "ods:type": {
+      "type": "string",
+      "description": "The DOI to the FDO type of the object",
+      "pattern": "^https://doi\\.org/[\\w.]+/[\\w.]+",
+      "examples": [
+        "https://doi.org/10.15468/1a2b3c"
+      ]
+    },
+    "ods:status": {
+      "type": "string",
+      "enum": [
+        "ods:Draft",
+        "ods:Active",
+        "ods:Tombstone"
+      ],
+      "description": "The status of the Digital Object",
+      "examples": [
+        "ods:Active"
+      ]
+    },
+    "schema:version": {
+      "type": "integer",
+      "minimum": 1,
+      "description": "Version of the Schema, https://schema.org/version",
+      "examples": [
+        1
+      ]
+    },
+    "schema:name": {
+      "type": "string",
+      "description": "Name of the Term, https://schema.org/name",
+      "examples": [
+        "ods:DigitalSpecimen"
+      ]
+    },
+    "skos:prefLabel": {
+      "type": "string",
+      "description": "The preferred lexical label for the term",
+      "examples": [
+        "Digital Specimen"
+      ]
+    },
+    "schema:dateCreated": {
+      "type": "string",
+      "description": "Timestamp of creation. Internally generated, https://schema.org/dateCreated",
+      "format": "date-time",
+      "examples": [
+        "2021-06-01T12:00:000Z"
+      ]
+    },
+    "schema:dateModified": {
+      "type": "string",
+      "description": "Timestamp of last modification. Internally generated, https://schema.org/dateModified",
+      "format": "date-time",
+      "examples": [
+        "2021-06-06T13:22:00.445Z"
+      ]
+    },
+    "schema:creator": {
+      "type": "object",
+      "description": "Contains an ods:Agent object",
+      "$ref": "https://schemas.dissco.tech/schemas/fdo-type/shared-model/0.3.0/agent.json",
+      "examples": [
+        {
+          "@id": "https://orcid.org/0000-0002-1825-0097",
+          "@type": "schema:Person",
+          "schema:name": "John Doe",
+          "schema:email": "john.doe@example.com"
+        }
+      ]
+    },
+     "schema:contributor": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "List of contributors",
+      "examples": [
+        ["Jane Smith", "Bob Johnson"]
+      ]
+    },
+    "skos:definition": {
+      "type": "string",
+      "description": "A definition of what the term represents.",
+      "examples": [
+        "A digital representation of a physical specimen"
+      ]
+    },
+   
+    "dcterms:licence": {
+      "type": "string",
+      "format": "uri",
+      "description": "License URI",
+      "enum": [
+        "https://creativecommons.org/licenses/by/4.0/",
+        "https://creativecommons.org/licenses/by-sa/4.0/",
+        "https://creativecommons.org/licenses/by-nc/4.0/",
+        "https://creativecommons.org/licenses/by-nc-sa/4.0/"
+      ],
+      "examples": [
+        "https://creativecommons.org/licenses/by/4.0/"
+      ]
+    },
+    "ods:belongsToClass": {
+      "type": "string",
+      "format": "uri",
+      "description": "URI of the class this term belongs to",
+      "examples": [
+        "https://schema.org/Thing"
+      ]
+    },
+    "ods:repeatable": {
+      "type": "boolean",
+      "description": "Indicates whether the term is repeatable",
+      "examples": [
+        true
+      ]
+    },
+    "ods:isRequired": {
+      "type": "boolean",
+      "description": "Indicates whether the term is required",
+      "examples": [
+        false
+      ]
+    },
+    "skos:broadMatch": {
+      "type": "string",
+      "format": "uri",
+      "description": "URI of a broader matching concept",
+      "examples": [
+        "http://rs.tdwg.org/dwc/terms/Occurrence"
+      ]
+    },
+    "owl:deprecated": {
+      "type": "boolean",
+      "description": "Indicates whether the term is deprecated",
+      "examples": [
+        false
+      ]
+    }
+  },
+  "required": [
+    "@type",
+    "ods:type",
+    "dcterms:identifier",
+    "schema:name",
+    "schema:dateCreated",
+    "schema:dateModified"
+  ]
+}


data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+    },
+    "ods:belongsToClass": {
+      "type": "string",
+      "format": "uri",
+      "description": "URI of the class this term belongs to",
+      "examples": [
+        "https://schema.org/Thing"
+      ]
+    },
+    "ods:repeatable": {
```
samleeflang, line 157: This and `ods:isRequired` feel like they should come from a generic ontology.
Maybe we could have a look at something like `owl:cardinality`

samleeflang, line 157: This doesn't comply to our `ods` naming convention.
For booleans this is `ods:isXxx`
So in this case it should be `ods:isRepeatable` in line with `ods:isRequired`

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+      "type": "boolean",
+      "description": "Indicates whether the term is repeatable",
+      "examples": [
+        true
+      ]
+    },
+    "ods:isRequired": {
```
samleeflang, line 164: These might also be tricky when we switch belongToClass to an array. It could then be required for one class and be optional for another.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+      "examples": [
+        "A digital representation of a physical specimen"
+      ]
+    },
+   
+    "dcterms:licence": {
+      "type": "string",
+      "format": "uri",
+      "description": "License URI",
+      "enum": [
```
samleeflang, line 139: Can't we make this a constant. All terms minted by openDS (us) would get the same license right? Or might there be different licenses used for different terms?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+    },
+    "skos:definition": {
+      "type": "string",
+      "description": "A definition of what the term represents.",
+      "examples": [
+        "A digital representation of a physical specimen"
+      ]
+    },
+   
+    "dcterms:licence": {
```
samleeflang, line 135: Mixing schema and dcterms. I would say stick with one or the other but mixing might be confusing. Most look to be coming from schema then I would suggest https://schema.org/license

samleeflang (commented):
Ok I think the start is there but I would make some changes:
- Don't mix schema.org and dcterms terms, there is a lot of overlap but I would say we pick one and stick with it
- Keep to naming convention
- BelongToClass probably needs to change but that impacts some other fields as well
-

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://schemas.dissco.tech/schemas/fdo-type/term/latest/term.json",
```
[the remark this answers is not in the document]
wouteraddink: I agree

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+      "examples": [
+        "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"
+      ]
+    },
+    "ods:type": {
+      "type": "string",
+      "description": "The DOI to the FDO type of the object",
+      "pattern": "^https://doi\\.org/[\\w.]+/[\\w.]+",
+      "examples": [
+        "https://doi.org/10.15468/1a2b3c"
```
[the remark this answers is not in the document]
wouteraddink: I think the FDO type should get a Handle not a DOI?
 regarding DOIs, I think I have seen a slightly different pattern used in code from Sou or Tom, including 10. This pattern is better, maybe we should use the same regex in all code. Note that the 21. examples are Handles not DOIs and may at some point not resolve through doi.org. Note also that DOIs expressed as DOI:10.1234/1234 are also valid DOIs.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+        "https://creativecommons.org/licenses/by/4.0/",
+        "https://creativecommons.org/licenses/by-sa/4.0/",
+        "https://creativecommons.org/licenses/by-nc/4.0/",
+        "https://creativecommons.org/licenses/by-nc-sa/4.0/"
+      ],
+      "examples": [
+        "https://creativecommons.org/licenses/by/4.0/"
+      ]
+    },
+    "ods:belongsToClass": {
```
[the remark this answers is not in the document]
wouteraddink: I would use the spdx urls for the machine readable version of the licenses, as we do in the FDO profiles also.

wouteraddink (changes requested):

data-model/fdo-type/term/term.json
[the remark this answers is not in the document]
sharifX: the "21." is coming from the epIC DTR.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+      "examples": [
+        {
+          "@id": "https://orcid.org/0000-0002-1825-0097",
+          "@type": "schema:Person",
+          "schema:name": "John Doe",
+          "schema:email": "john.doe@example.com"
+        }
+      ]
+    },
+     "schema:contributor": {
```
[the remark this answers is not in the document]
sharifX: ok. yes. the definition mentions "secondary contributor".

data-model/fdo-type/term/term.json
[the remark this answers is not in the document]
sharifX: I got the "belongsToClass" idea from ltc. they only have one term but yes, we can make this an array.

data-model/fdo-type/term/term.json
sharifX: ok. yes, I was struggling to find one for this. owl:cardinality seems like a good option.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
...
+    },
+    "dcterms:identifier": {
+      "type": "string",
+      "description": "Handle of the Term",
+      "pattern": "^https://hdl\\.handle\\.net/[\\w.]+/(.){3}-(.){3}-(.){3}",
+      "examples": [
+        "https://hdl.handle.net/20.5000.1025/XXX-XXX-XXX"
+      ]
+    },
+    "ods:type": {
```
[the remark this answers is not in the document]
sharifX: updated. in the new schema

data-model/fdo-type/term/term.json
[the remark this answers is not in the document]
sharifX: kept the example same now as the other digital objects.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,195 @@
+{
```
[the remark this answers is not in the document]
sharifX: this in the json schema?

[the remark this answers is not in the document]
samleeflang: Yes certainly, every term has examples in the schema

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,168 @@
...
+    },
+    "@type": {
+      "type": "string",
+      "description": "The type of the object, in this case ods:Term",
+      "const": "ods:Term",
+      "examples": [
+        "ods:Term"
+      ]
+    },
+    "rdf:type": {
```
samleeflang, line 24: Isn't the `@type` the same as `rdf:type` ? But here they have different values, how would this work?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,168 @@
...
+        "ods:Draft",
+        "ods:Active",
+        "ods:Tombstone"
+      ],
+      "description": "The status of the Digital Object",
+      "examples": [
+        "ods:Active"
+      ]
+    },
+    "ods:version": {
```
samleeflang, line 67: Do we version the terms itself, at the moment we only version the DigitalObject. We could probably do this but don't we need more provenance attributes than? Such as `owl:versionInfo` or `owl:previousVersion` `skos:historyNote`, `dcterms:isVersionOf`, etc...?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,168 @@
...
+      "type": "string",
+      "description": "The timestamp that the object was last changed, following the ISO Date Time Format yyyy-MM-dd'T'HH:mm:ss.SSSXXX",
+      "examples": [
+        "2023-10-02T12:31:34.806Z"
+      ]
+    },
+    "dcterms:creator": {
+      "type": "object",
+      "description": "Contains information about the creator/proposer/requester of the term. This is an ods:Agent object",
+      "$ref": "https://schemas.dissco.tech/schemas/fdo-type/shared-model/0.4.0/agent.json",
```
samleeflang, line 107: I am not sure about the use of a openDS object here. The example would look a bit differently as the orcid should also be there as an identifier.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,168 @@
...
+    "dcterms:modified": {
+      "type": "string",
+      "description": "The timestamp that the object was last changed, following the ISO Date Time Format yyyy-MM-dd'T'HH:mm:ss.SSSXXX",
+      "examples": [
+        "2023-10-02T12:31:34.806Z"
+      ]
+    },
+    "dcterms:creator": {
+      "type": "object",
+      "description": "Contains information about the creator/proposer/requester of the term. This is an ods:Agent object",
```
samleeflang, line 106: These are three different roles but there is just room for a single creator. Can it occur that there are multiple creators or both a requester and a creator?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,168 @@
...
+      "description": "A definition of what the term represents.",
+      "examples": [
+        "A digital representation of a physical specimen"
+      ]
+    },
+    "dcterms:licence": {
+      "const": "https://spdx.org/licenses/CC0-1.0.json",
+      "description": "The licence for the term, required to be always public domain"
+    },
+    "ods:belongsToClass": {
```
samleeflang, line 128: Wondering if `rdfs:range` does not cover this? This doesn't feel like we should be the ones creating this term

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,168 @@
...
+      "items": {
+        "type": "string",
+        "format": "uri"
+      },
+      "description": "URIs of the classes this term belongs to",
+      "examples": [
+        ["https://terms.dissco.tech/digital-specimen-terms#EntityRelationship", "http://purl.org/dc/dcmitype/Collection"]
+      ]
+    },
+    "owl:cardinality": {
```
samleeflang, line 139: Do we need all three, if we have the other two that would be enough right? Is there a use case when those two won't suffice?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,168 @@
...
+    "owl:deprecated": {
+      "type": "boolean",
+      "description": "Indicates whether the term is deprecated",
+      "examples": [
+        false
+      ]
+    }
+  },
+  "required": [
+    "@type",
```
samleeflang, line 163: Still missing `@id` here and I think more terms can/should be mandatory

samleeflang (commented):
Might be best to present this in the exploration meeting, I have difficulty grasping how this would function in the landscape? What will the identifiers look like, do we need changes to the terms (for example add version)? How does this fit with semantic best practise (we use owl terms but don't implement owl)?

data-model/fdo-type/term/term.json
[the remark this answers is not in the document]
sharifX: removed contributor for now.

data-model/fdo-type/term/term.json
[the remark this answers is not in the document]
sharifX: ok. will add this.

data-model/fdo-type/term/term.json
sharifX: check tdwg and add previous version

sharifX: add Notes (for decision context)

sharifX:
updated schema 
changes in this version: 
- updated Regex for the Handle 
- removed rdf:Type 
- used ods:hasAgent instead of dcterms:creator 
- added skos:note and skos:example 
- removed owl cardinality. we use json schema required for this 
-  added hasVersion for more provenance metadata 
- used rdfs:range instead of new term belongsToClass 
- added language in prefLabel

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://schemas.dissco.tech/schemas/fdo-type/term/0.1.0/term.json",
+  "title": "DiSSCo openDS Term Schema",
+  "description": "A schema for defining standard terms in openDS.",
+  "type": "object",
+  "properties": {
+    "@id": {
```
samleeflang, line 8: For other object we also have a @type which is the readable version and for this I presume it would be "ods:Term"? Or is there a reason this is not included

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+      "type": "string",
+      "description": "The unique identifier (handle) of the term object",
+      "pattern": "^https://hdl\\.handle\\.net/20\\.500\\.1025/([^/]+)$",
+      "examples": [
+        "https://hdl.handle.net/20.500.1025/DigitalSpecimen"
+      ]
+    },
+    "ods:fdoType": {
```
samleeflang, line 16: So if I understand correctly we will get an ods term describing itself. So we will get an ods:fdoType object (https://hdl.handle.net/20.500.1025/fdoType) which has then a property which is defined by itself? Is this correct use of the terms?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+      ]
+    },
+    "dcterms:modified": {
+      "type": "string",
+      "description": "The timestamp that the object was last changed, following the ISO Date Time Format yyyy-MM-dd'T'HH:mm:ss.SSSXXX",
+      "examples": [
+        "2023-10-02T12:31:34.806Z"
+      ]
+    },
+
```
samleeflang, line 113: Unnecessary whitespace

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+    },
+    "dcterms:modified": {
+      "type": "string",
+      "description": "The timestamp that the object was last changed, following the ISO Date Time Format yyyy-MM-dd'T'HH:mm:ss.SSSXXX",
+      "examples": [
+        "2023-10-02T12:31:34.806Z"
+      ]
+    },
+  
+    "dcterms:hasVersion": {
```
samleeflang, line 114: How does this work together with "ods:version"

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+      "items": {
+        "type": "string",
+        "format": "uri"
+      },
+      "description": "URIs of the classes this term ranges to or belongs to",
+      "examples": [
+        ["https://terms.dissco.tech/digital-specimen-terms#EntityRelationship", "http://purl.org/dc/dcmitype/Collection"]
+      ]
+    },
+    "owl:deprecated": {
```
samleeflang, line 132: Shouldn't this be always present and therefore required?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+    },
+  
+    "dcterms:hasVersion": {
+      "type": "string",
+      "description": "A related resource that is a version, edition, or adaptation of the described resource.",
+      "examples": [
+        "1.0.1"
+      ]
+    },
+    "rdfs:range": {
```
samleeflang, line 121: Shouldn't this be always present and therefore required? When is a term not part of a class?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+      }
+    },
+    "dcterms:title": {
+      "type": "string",
+      "description": "A Name given to the resource, in this case openDS term name",
+      "examples": [
+        "ods:DigitalSpecimen"
+      ]
+    },
+    "skos:prefLabel": {
```
samleeflang, line 60: Shouldn't this be always present and therefore required?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+      "type": "string",
+      "description": "The preferred lexical label for the term",
+      "examples": [
+        "Digital Specimen"
+      ],
+      "@language": {
+        "type": "string"
+      }
+    },
+    "skos:definition": {
```
samleeflang, line 70: Shouldn't this be always present and therefore required?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+    },
+    "ods:version": {
+      "type": "integer",
+      "description": "The version of the object, each change generates a new version. The version starts at 1 and each change will increment the version number with 1",
+      "minimum": 1,
+      "examples": [
+        1
+      ]
+    },
+ "ods:hasAgents": {
```
samleeflang, line 45: Shouldn't this be always present and therefore required? Shouldn't we always store a creator and a modifier?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+    "ods:fdoType": {
+      "type": "string",
+      "description": "The DOI to the FDO type of the object",
+      "pattern": "^https://doi\\.org/[\\w\\.]+/[\\w\\.]+",
+      "examples": [
+        "https://doi.org/21.T11148/bbad8c4e101e8af01115",
+        "https://doi.org/21.T11148/894b1e6cad57e921764e"
+      ]
+    },
+    "ods:status": {
```
samleeflang, line 25: Shouldn't this be always present and therefore required?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,148 @@
...
+    },
+    "owl:deprecated": {
+      "type": "boolean",
+      "description": "Indicates whether the term is deprecated",
+      "examples": [
+        false
+      ]
+    }
+  },
+  "required": [
```
samleeflang, line 140: Missing a couple of required attributes here

samleeflang (commented):
Couple comments. General points:
- Can a term reference itself? We use odsTerms here which are described by this object. So the object is using a term which is defined by itself. Not sure if that is an issue or not.
- I think we need a lot more required attributes

data-model/fdo-type/term/term.json
sharifX: Yes, this is common in RDF/ontology and data typing world. self referencing/recursive definition. 
For example in RDF schema (https://www.w3.org/2000/01/rdf-schema) 
rdfs:Class is the class of all classes, and it is itself an instance of rdfs:Class.
```
rdfs:Class a rdfs:Class ;
	rdfs:isDefinedBy <http://www.w3.org/2000/01/rdf-schema#> ;
	rdfs:label "Class" ;
	rdfs:comment "The class of classes." ;
	rdfs:subClassOf rdfs:Resource .
```

This also supports FDO Forum recommendation. (https://zenodo.org/records/14614908 page 7). 
exact quote from the pdf in the link: 
`"attribute definitions are capable to also describe definitions of attribute definitions."`

and yes, there could be performance issues related to resolution and serialisation/deserialisation. but that could be solved other ways.

data-model/fdo-type/term/term.json
sharifX: as discussed in the meeting a few weeks ago, the term schema will have hasAgents required and call the object. we do not need to list hasRoles in term schema separately.

data-model/fdo-type/term/term.json
sharifX: yes, added

data-model/fdo-type/term/term.json
sharifX: yes, added

data-model/fdo-type/term/term.json
sharifX: added @type similar to the others.

data-model/fdo-type/term/term.json
sharifX: we are using rdfs:range. updated.

data-model/fdo-type/term/examples/ods:topicDiscipline-example.json
```diff
@@ -0,0 +1,23 @@
+{
+  "@id": "https://hdl.handle.net/20.500.1025/topicDiscipline",
+  "@type": "ods:Term",
+  "ods:fdoType": "https://doi.org/21.T11148/bbad8c4e101e8af01115",
+  "ods:status": "ods:Active",
+  "ods:version": 1,
+  "ods:hasAgents": [],
```
samleeflang, line 7: This should have a creator and/or modifier

samleeflang (commented):
Have some comments, mainly about the examples.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,184 @@
...
+        "ods:DigitalSpecimen"
+      ]
+    },
+    "skos:prefLabel": {
+      "type": "string",
+      "description": "The preferred lexical label for the term",
+      "examples": [
+        "Digital Specimen"
+      ],
+      "@language": {
```
samleeflang: This isn't a valid JsonSchema property

data-model/fdo-type/term/term.json
samleeflang: This is still here. I still have the same question.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,184 @@
...
+    "ods:fdoType",
+    "dcterms:created",
+    "dcterms:modified", 
+    "dcterms:title",
+    "owl:deprecated",
+    "skos:prefLabel",
+    "skos:definition",
+    "ods:hasAgents",
+    "ods:status",  
+    "rdfs:range"
```
samleeflang: Would range always be there? If I got a ods:DigitalSpecimen which is the class than it won't have a range correct?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,184 @@
...
+      "type": "string",
+      "description": "A note field to capture contexts and other information around the decision.",
+      "examples": [
+        "This term was discussed in projectX"
+      ],
+      "@language": {
+        "type": "string"
+      }
+    },
+  "skos:example": {
```
samleeflang: Should we make this a required field?

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,184 @@
...
+    },
+    "@type": {
+      "type": "string",
+      "const": "ods:Term",
+      "description": "The type of the digital object, in this case a ods:Term"
+    },
+ "ods:fdoType": {
+      "type": "string",
+      "description": "The Handle to the FDO type of the object",
+      "pattern": "^https:\/\/hdl\\.handle\\.net\/[\\w.]+\/(.){3}-(.){3}-(.){3}",
```
samleeflang: This is an incorrect pattern. Even the example doesn't validate against this regex.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,184 @@
...
+      "examples": [
+        "https://hdl.handle.net/21.T11148/2ac65a933b7a0361b651"
```
samleeflang: This is still the VirtualCollection, better to point to the FDOType for the Term here

samleeflang (commented):
Check the comments, but there are still some basic issue with this PR.
Please validate that the properties you use to describe are valid JSON Schema properties. You can just check the standard https://json-schema.org/.
Validate the examples against the schema.
This can easily be done through something like https://www.jsonschemavalidator.net/ 
You will immediately see there are issue with the validation of the examples.

data-model/fdo-type/term/term.json
```diff
@@ -0,0 +1,184 @@
...
+      "type": "string",
+      "description": "The preferred lexical label for the term",
+      "examples": [
+        "Digital Specimen"
+      ],
+      "@language": {
+        "type": "string"
+      }
+    },
+    "skos:member": {
```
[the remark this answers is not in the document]
sharifX: ah yes, this can be removed. ods:hasControlledvalue does the same purpose justification is that the proper use of skos:member is related to skos:Collections. we are not using that. i added some notes in DEX-106.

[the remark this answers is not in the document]
sharifX: and formatting fixed.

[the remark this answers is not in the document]
sharifX: this can be removed. ods:hasControlledvalue does the same purpose justification is that the proper use of skos:member is related to skos:Collections. we are not using that. i added some notes in DEX-106. and yes. each term will be a dissco vocab. ods:topicDiscipline example has this.

PR #198, opened 2024-09-18
Change env and secret for MAS

samleeflang:
Discussed over slack:
- add namespace to terms
- align object name with term name
- fix required fields
- move to shared objects to enable other objects to reuse it 
- added examples


southeo (approved):
:lock:

PR #200, opened 2024-09-24
split tombstone request and attributes

southeo:
- publish 1.0.0 of tombstone schema
- split attributes from whole request - this makes it easier to generate java objects later


samleeflang (approved):
:takeout_box:

TomDijkema (approved):
:woman_cook:

PR #238, opened 2025-06-25
DwC-DP -> openDS mapping

samleeflang:
Not sure what the expectation is for the field mapping.
I thought this repo would be a nice place to store it.
The csv consist with the most basic things possible:
DwC-DP term, openDS term
For most this works fine.
I also added some short remarks with how the mapping works.

I got more complicated files as well but these are more for internal use/discussion
https://docs.google.com/spreadsheets/d/1tUUCL363OTQjPzBq6fa-gmAKCfOH5AGRlNbNnY9LOp0/edit?usp=sharing


data-mapping/dwc-dp/material/README.md
```diff
@@ -0,0 +1,19 @@
+# Material Mapping to DwC-DP
+
+Most fields can be mapped with a one to one mapping to the DwC-DP.
+A couple of new fields are not present in openDS which might require some discussion.
+
+We ignore the taxonomical fields in the material file, as we use the DwC-DP Identification file for this.
```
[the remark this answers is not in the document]
southeo: maybe in a comment column?

data-mapping/dwc-dp/media/README.md
```diff
@@ -0,0 +1,6 @@
+# Media to DwC-DP
+
+For media we lack a couple fields which are not present in openDS.
+We could access there relevance and potentially add them in the future.
```
southeo, line 4: Do you mean, "We could assess their relevance" here?

data-mapping/dwc-dp/material/material-mapping.csv
```diff
@@ -0,0 +1,47 @@
...
+materialEntityType,
+digitalSpecimenID,$.@id
+institutionCode,$.ods:organisationCode
+institutionID,$.ods:organisationID
+ownerInstitutionCode,$.ods:ownerOrganisationCode
+ownerInstitutionID,
+collectionCode,$.dwc:collectionCode
+collectionID,$.dwc:collectionID
+catalogNumber,"Filled based on an Identifier if the Identifier title is ""dwc:catalogNumber"""
+otherCatalogNumbers,Not filled we use material-identifier for otherCatalogNumbers
```
southeo, line 13: Add quotes?

data-mapping/dwc-dp/relationship/README.md
```diff
@@ -0,0 +1,8 @@
+# Relationship to DwC-DP
+
+There are differences in how openDS and DwC-DP handle relationships.
+openDS sees every link to an external resource as an entity relationship.
+This means that url's present in the metadata are often duplicated in the relationship.
```
southeo, line 5: urls

southeo (commented):
I think what this is missing is a readme with information about your approach to dissco data mappings.

- When there is no equivalent term in openDS, we leave that term blank but include it in the mapping anyway
- We use json path notation to indicate what part of openDS the term is from. Where arrays are present, we use [*] (unless specified, e.g. in the event)
- When a value is a constant (aka hardcoded)
 - What we do when multiple terms exist in openDS for one term in the corresponding model (I'd suggest piping these terms)

I'd also suggest having a comment column. That way, the openDS column can be machine actionable

data-mapping/dwc-dp/agent/README.md
```diff
@@ -0,0 +1,7 @@
+# Agent Mapping to DwC-DP
+
+The structure for agents is a bit different in openDS and DwC-DP.
```
samleeflang, line 3: When DiSSCo empty it will not be in the DwC-DP

data-mapping/dwc-dp/identification/identification-mapping.csv
```diff
@@ -0,0 +1,20 @@
...
+taxonFormula,$.ods:hasIdentifications[*].dwc:taxonFormula
+typeStatus,$.ods:hasIdentifications[*].dwc:typeStatus
+typeDesignationType,
+typifiedName,
+identifiedBy,$.ods:hasIdentifications[*].ods:hasAgents
+identifiedByID,$.ods:hasIdentifications[*].ods:hasAgents
+dateIdentified,$.ods:hasIdentifications[*].dwc:dateIdentified
+identificationReferences,
+taxonAssignmentMethod,
+identificationVerificationStatus,$.ods:hasIdentifications[*].dwc:identificationVerificationStatus
```
wouteraddink, line 19: In openDS I cannot find $.ods:hasIdentifications[*].dwc:identificationVerificationStatus, but we have ods:isVerifiedIdentification. Is this an error?

wouteraddink (changes requested):
please look at my comment for identificationVerificationStatus, otherwise it looks fine for me. 

I have some suggestions for improvement:
1. I would add a readme for the whole mapping that shortly explains the used syntax for the source path, although it is clear for me. Maybe you could use an example like:

agentRoleOrder: $.ods:hasEvents[0].ods:hasAgents[*].ods:hasRoles[*].schema:position

Explanation:
Darwin Core Data Package expects agentRoleOrder to be a flattened field indicating the position/order of the agent in the event.
ods:hasEvents[0]: First event related to the specimen.
ods:hasAgents[*]: All agents involved in that event.
ods:hasRoles[*]: Each agent may have multiple roles.
schema:position: A role's position (e.g., 1st, 2nd in sequence).

2. It would be useful to add headers to the CSV file: **target_field, source_path**. I would add a few more columns: **transformation, constant**,  as you sometimes use a constant and you may also use transformations (although I think I did not see any), e.g. lowercase.
I would also add a column for **comments**.
3. In the future it may be a good idea to use the csv files directly in the mapping, that would guarantee they are always up to date. You than would probably need an additional "group" column to indicate groupings as you sometimes need to combine values from multiple terms into one (or split values). Drawback is that the csv becomes harder to read than.
4. I agree with Sou that it would be good to also include openDS terms that could not be mapped, or mention them in the readme.

PR #253, opened 2026-04-20
remove topic name from mas request

southeo:
[jira](https://naturalis.atlassian.net/browse/DD-3081)

**Removing the topic name from mas request**

The MAS topic name is currently internally managed, unless a user overwrites it in a request. By default, the topic is the lower-case suffix of the PID, but it can be overwritten by a user. 

There is a case where the Orchestration front end can overwrite the topic to be null, and it is not caught. 

Instead of adding a check to ensure the topic isn’t null in a request, the topic should be omitted from the request entirely, and be fully internally managed. 

**Accepted annotation metadata**
Also including accepted annotation metadata in annotation request so that we can accept annotations.


samleeflang (approved):
🍞

